PR #2, opened 2019-07-17
Fix content examples for OAS2

XVincentX:
In the current implementation every example gets copied in every content — but they should not. In OAS2 the *key* of the example is the *mimeType* of the example itself, and they should go in their relative content.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "yalc.watch": "nodemon --watch 'src/**/*' --exec 'yarn yalc'"
   },
   "dependencies": {
-    "@stoplight/json": "^2.3.2",
+    "@stoplight/json": "^3.0.1",
     "@stoplight/types": "^9.1.2",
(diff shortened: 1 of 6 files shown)


XVincentX:
@marbemac Can you check this out?

marbemac:
Bah, you beat me to the merge and the merge commit message is not in correct format - won't release.

XVincentX:
Arghm I’l create another PR

marbemac:
Is this basically lossy now? If you have an example key in oas2 that doesn't exactly match a produces, is it thrown away?

Perhaps we should instead just toss it into `contents[0].examples`? This is a small change that would help us maintain support for the random example pattern from oas2 yesteryear in mocking and docs, while people migrate. Our new tooling will still not allow random examples in oas2 moving forward, so it's basically "deprecated" instead of "no more support at all".

cc @XVincentX @philsturgeon

PR #5, opened 2019-07-17
fix: restore sl-scripts build

XVincentX:
Closes #1 

I wanted to get rid of the 3 build files but there are file copies here and there required in the build process I can't take care right now


marbemac:
could you update scripts package back to `^5.1`?

PR #8, opened 2019-07-22
fix: Do not discard foreign values

### LINKED ISSUE #7: Keep invalid (not content type) examples
Whilst the OAS2 spec does say example names SHOULD be a mime type, and it does say they MUST be in produces, we need to (for now at least) leave them intact for class/next docs/prism users who are used to being able to do this.

Solution:

IF an oas2 example does match a produces, push it into `contents[0].examples`. It would be a small change that would help us maintain support for the random example pattern from oas2 yesteryear in mocking and docs, while people migrate. Our new tooling (e.g: Studio) will still not allow random examples in oas2 moving forward unless they modify in the code view, so it's basically "deprecated" instead of "no more support at all".

**Additional Context**

A followup to https://github.com/stoplightio/http-spec/pull/2

We can add a rule to Spectral to check for examples not in produces, so users get more insight into the problem, and between that and Studio giving them a nudge, it might just convince them to upgrade to OAS3.

_Originally posted by @marbemac in https://github.com/stoplightio/http-spec/pull/2#issuecomment-513281772_

XVincentX:
This PR will put the "foreign examples" in the first content — or create a new one in case there's no content.

I also found an incorrect behavior that I fixed — fundamentally an (empty) content was produced per each `produce` entry even thought an operation would *not* produce a content of such type.

With this change we have:

1. Smaller output size
2. Clearly differentiate between
  * `produce` for the current document is provided but the operation does not produce such content
  * `produce` is provided and the current operation provides an *explicit empty content*

Closes #7


package.json
```diff
@@ -59,8 +54,8 @@
     "typescript": "3.5.3"
   },
   "lint-staged": {
-    "*.{ts,tsx}$": [
-      "yarn lint.fix",
+    "src/**/*.ts": [
+      "tslint -p tsconfig.build.json",
```
marbemac: Porque?

src/oas2/transformers/__tests__/responses.test.ts
```diff
@@ -55,7 +56,7 @@ describe('responses', () => {
     ).toMatchSnapshot();
   });
 
-  test('should translate to response w/o examples', () => {
+  test.only('should translate to response w/o examples', () => {
```
marbemac: Remove. @philsturgeon might have been a little hasty of a review ;).

package.json
XVincentX: Básicamente en la manera precedente el sistema de linting estaba comprobando el proyecto por entero aunque se cambiaba un file soltero.

:trollface:

XVincentX: All right — basically with that command apparently the whole project is always checked. With this new command, only the effective changed files will be checked

XVincentX:
@marbemac I understand what you mean.

What I understood is that `produces` claims what an operation *could* produce, not what *must* produce. Therefore there might be operations that return only JSON, although the document is claiming, on the top, to return JSON and XML.

If we leave the things as it is, we'd basically claim that all the operation can produce whatever is in the produce array, no matter what.

With this change, this would not happen anymore.

I'm ok with keeping it though, I thought it was a bug.

marbemac:
> I'm ok with keeping it though, I thought it was a bug.

I think we should keep it. Otherwise folks will wonder why we're not indicating in the generated docs that xyz can produce json. Best not to be lossy in our conversion from openapi -> IHttpOperation.

XVincentX:
@marbemac You're right. I was confused and thought that a content is an example — which is not true.

A content is something that *might* have an example; the fact that there's no example does not mean that the API cannot produce such content type representation. I've restored the behaviour in 98f8f79

Check it one more time please!

XVincentX:
@marbemac can you help bring this over the finish line?

PR #9, opened 2019-07-26
Bug/prism/483/form-data-validation

chris-miaskowski:
https://github.com/stoplightio/prism/issues/483
 https://github.com/stoplightio/prism/pull/482

## What was the issue

Form Data in OAS2 is much different than in OAS3. E.g. in OAS3 everything is a schema. In OAS2 on the other hand form data is a parameter similar to path or query param. We tranformed OAS2 form data parameter incorrectly by throwing everything in `encoding`. This way the `schema` was incorrecly built and Prism could validate it. 

## Changes

- refactor body form-data validation
- update tests


src/oas2/transformers/params.ts
```diff
@@ -81,41 +82,84 @@ export function translateToBodyParameter(body: BodyParameter, consumes: string[]
 }
 
 export function translateFromFormDataParameter(
-  form: FormDataParameter,
-  requestBody: IHttpOperationRequestBody | null | undefined,
+  parameters: FormDataParameter[],
   consumes: string[],
 ): IHttpOperationRequestBody {
-  const baseContent = buildContentFromParameter(form);
```
XVincentX: Apparently you've changed this function quite a lot. According to the code this was transforming a *single* parameter, now it's transforming all the parameters all together. Was there any particular reason for this?

src/oas2/transformers/params.ts
```diff
@@ -81,41 +82,84 @@ export function translateToBodyParameter(body: BodyParameter, consumes: string[]
...
+          content.encodings.push(encoding);
+        }
+      }
+    });
+    return body;
+  }, finalBody);
+}
 
-    bodyContent.encodings = (bodyContent.encodings || []).concat(encoding);
+function buildEncoding(parameter: FormDataParameter, content: IMediaTypeContent): IHttpEncoding | null {
```
XVincentX: The `content` property seems not used, can we remove it?

src/oas2/transformers/params.ts
```diff
@@ -157,11 +201,9 @@ function buildContentFromParameter(
...
   );
 
-  return {
-    schema: pickBy(
-      defaults<Schema, Schema>(schema, {
-        minLength: 'allowEmptyValue' in param && param.allowEmptyValue ? 1 : undefined,
-      }),
-    ),
-  };
+  if ('allowEmptyValue' in param && param.allowEmptyValue) {
```
XVincentX: Yeah, I admit the readability of this line of code is way better than the previos one 👍

src/oas2/transformers/params.ts
```diff
@@ -81,41 +82,84 @@ export function translateToBodyParameter(body: BodyParameter, consumes: string[]
...
-      property: form.name,
-      style: HttpParamStyles.Form,
-    };
+  return parameters.reduce((body, parameter) => {
+    const { schema } = buildSchemaForParameter(parameter);
+    (body.contents || []).forEach(content => {
+      // workaround... JSONSchema4 does not support `allowEmptyValue`
+      if ('allowEmptyValue' in parameter) {
+        // @ts-ignore
+        schema.allowEmptyValue = parameter.allowEmptyValue;
```
XVincentX: You can use `Object.assign` to avoid to shutdown the type checker on this line

```suggestion
        Object.assign(schema, { allowEmptyValue: parameter.allowEmptyValue });
```

src/oas2/transformers/params.ts
```diff
@@ -81,41 +82,84 @@ export function translateToBodyParameter(body: BodyParameter, consumes: string[]
...
+      }
+      set(content, `schema.properties.${parameter.name}`, schema);
+      if (parameter.required) {
+        const requiredIndex = get(content, 'schema.required', []).length;
```
XVincentX:
```suggestion
        const requiredIndex = get(content, 'schema.required.length', 0);
```

src/oas2/transformers/request.ts
```diff
@@ -11,21 +11,31 @@ import {
...
-      const headerParameter = translateToHeaderParam(parameter);
-      request.headers = (request.headers || []).concat(headerParameter);
-    }
-    return request;
-  }, {});
+  const bodyParameters = parameters.filter(parameter => isBodyParameter(parameter)) as BodyParameter[];
+  const formDataParameters = parameters.filter(parameter => isFormDataParameter(parameter)) as FormDataParameter[];
+  const request: IHttpOperationRequest = {};
+
+  // if 'body' and 'form data' defined prefer 'body'
```
XVincentX: Yeah I think this is a good choice. The spec does not really mandate that a parameter with `formData` excludes `body` and viceversa.

src/oas2/transformers/params.ts
```diff
@@ -19,7 +20,7 @@ import {
 } from 'swagger-schema-official';
 
 function chooseQueryParameterStyle(
```
XVincentX: This function is not taking a `FormDataParameter` as well — so the name is probably not ok anymore; in any case this is only used in the `QueryParameter` path so I'm not too sure why this change is here

src/oas2/transformers/params.ts
chris-miaskowski: Hey, good question. Several reasons:
1. The original function horrible - it "kind of mutated inputs"...
2. An array of parameters is reduced into a single schema - it previously took each param individually and "existing body" in order to "mutate" that existing body. That was redundant and confusing.
3. It didn't build `encodings` correctly.

src/oas2/transformers/params.ts
```diff
@@ -81,41 +82,84 @@ export function translateToBodyParameter(body: BodyParameter, consumes: string[]
...
+        style: HttpParamStyles.Form,
+        explode: true,
+      };
+    case 'ssv':
+      return {
+        property: parameter.name,
+        style: HttpParamStyles.SpaceDelimited,
+        explode: false,
+      };
+    // TODO: TSV is not supported in OAS3, no compatibility
```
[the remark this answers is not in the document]
chris-miaskowski: Will remove the comment and create an issue. The point I was trying to make is to leave a reminder that our current types do not allow for TSV. We probably need to make our types more flexible.

src/oas2/transformers/request.ts
```diff
@@ -11,21 +11,31 @@ import {
...
-    } else if (isPathParameter(parameter)) {
-      const pathParameter = translateToPathParameter(parameter);
-      request.path = (request.path || []).concat(pathParameter);
-    } else if (isHeaderParameter(parameter)) {
-      const headerParameter = translateToHeaderParam(parameter);
-      request.headers = (request.headers || []).concat(headerParameter);
-    }
-    return request;
-  }, {});
+  const bodyParameters = parameters.filter(parameter => isBodyParameter(parameter)) as BodyParameter[];
```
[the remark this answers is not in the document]
chris-miaskowski: Yup, when passed function directly it works like a charm.

src/oas2/transformers/params.ts
chris-miaskowski: Oh, overshight! Removing!

src/oas2/transformers/params.ts
XVincentX: In that case we should change the name

PR #11, opened 2019-07-30
fix: Handle nullable property in OAS3

XVincentX:
* Nullable properties were not converted in a regular JSONSchema thing.

This is for https://github.com/stoplightio/prism/issues/486


src/oas2/transformers/__tests__/params.test.ts
```diff
@@ -162,20 +162,21 @@ describe('params.translator', () => {
...
               maxItems: 10,
               minItems: 1,
+              minLength: 1,
               type: 'array',
             },
             int: {
               description: 'desc',
               maximum: 3,
               minimum: 0,
+              minLength: 1,
```
marbemac: Maybe I'm not understanding this PR, but `minLength` is a string validation not an integer/number validation.

src/oas3/transformers/content.ts
```diff
@@ -103,3 +104,11 @@ const transformExamples = (source: MediaTypeObject | HeaderObject) => (key: stri
     key,
   };
 };
+
+const transformSchema = (schema: SchemaObject): JSONSchema4 | JSONSchema6 | JSONSchema7 => {
```
marbemac: Clone deep is not necessary, suggest something like this maybe?

```ts
const transformSchema = (schema: SchemaObject): JSONSchema4 | JSONSchema6 | JSONSchema7 => {
   if (schema && schema.nullable) {
     return {
       ...schema,
       type: [schema.type, 'null']
     }
   }

   return schema;
};
```

src/oas2/transformers/params.ts
```diff
@@ -192,6 +192,8 @@ function buildSchemaForParameter(
   );
 
   if ('allowEmptyValue' in param && param.allowEmptyValue) {
+    delete schema.minLength;
```
marbemac: What is this fixing?

XVincentX: Whoops, my bad — I thought I updated the PR text.

Basically if `allowEmptyValue` is true, the `minLenght` property is "unset" so that empty strings are allowed.

src/oas3/transformers/content.ts
XVincentX: Done!

src/oas2/transformers/__tests__/params.test.ts
XVincentX: Yeah, this test is likely wrong. I'll look into.

XVincentX:
@marbemac I think I accidentally mixed two PRs together. I've removed the foreign code.

PR #12, opened 2019-07-30
fix: nested properties

XVincentX:
Convert to JSON Schema the JSONish one that OAS3 provides.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "@types/swagger-schema-official": "~2.0.18",
     "@types/urijs": "~1.19.1",
     "lodash": "^4.17.11",
+    "openapi-schema-to-json-schema": "2.2.0",
     "openapi3-ts": "~1.3.0",
     "urijs": "~1.19.1"
   },
--- a/src/oas3/transformers/__tests__/__snapshots__/content.test.ts.snap
+++ b/src/oas3/transformers/__tests__/__snapshots__/content.test.ts.snap
@@ -58,7 +58,9 @@ Object {
     },
   ],
   "mediaType": "mediaType",
-  "schema": Object {},
+  "schema": Object {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+  },
 }
 `;
 
@@ -107,7 +109,9 @@ Object {
     },
   ],
   "mediaType": "mediaType",
-  "schema": Object {},
+  "schema": Object {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+  },
 }
 `;
 
@@ -121,7 +125,9 @@ Object {
     },
   ],
   "mediaType": "mediaType",
-  "schema": Object {},
+  "schema": Object {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+  },
 }
 `;
 
@@ -135,6 +141,8 @@ Object {
     },
   ],
   "mediaType": "mediaType",
-  "schema": Object {},
+  "schema": Object {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+  },
 }
 `;
--- a/src/oas3/transformers/content.ts
+++ b/src/oas3/transformers/content.ts
@@ -106,11 +106,24 @@ const transformExamples = (source: MediaTypeObject | HeaderObject) => (key: stri
 };
 
 const transformSchema = (schema: SchemaObject) => {
-  if (schema && schema.nullable) {
-    return {
-      ...schema,
-      type: [schema.type, 'null'],
-    };
+  if (schema) {
+    if (schema.nullable)
+      return {
+        ...schema,
+        type: [schema.type, 'null'],
+      };
+
+    if (schema.properties) {
+      Object.keys(schema.properties).forEach(prop => {
+        schema.properties![prop] = transformSchema(schema.properties![prop]) as SchemaObject;
+      });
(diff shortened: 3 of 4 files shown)


XVincentX:
@marbemac Better to use a library doing this for us. Please check it out again.

marbemac:
Why better to use a library? It's a very small amount of code.

This is breaking downstream consumers, because is too "pure" / not very fault tolerant. For example, it throws errors if the spec is not what it expects: https://github.com/wework/json-schema-to-openapi-schema/blob/master/index.js#L48. 

We don't have this luxury, lots of specs are not perfect. For example, this is what happens loading the sendgrid spec inside of studio with this library:

Screen Shot 2019-08-02 at 11 11 12 PM

It's because graphite is expecting http-spec to be fault tolerant - this is a guiding principle for much of our tooling. We need to do our best to mostly work under less than ideal circumstances, and let still the user know that things are not perfect.

@XVincentX @philsturgeon I suggest that we revert this.

philsturgeon:
@marbemac you're looking at the wrong library. 😅

There is [mikunn/openapi-schema-to-json-schema](https://github.com/mikunn/openapi-schema-to-json-schema) and [wework/json-schema-to-openapi-schema](https://github.com/wework/json-schema-to-openapi-schema). 

We used mikunn/openapi-schema-to-json-schema in this PR to convert our OpenAPI into JSON Schema, but you're looking at wework/json-schema-to-openapi-schema.

So why use a library instead of the small amount of code? Well, the small amount of code doesn't do enough:

```ts
const transformSchema = (schema: SchemaObject) => {
  if (schema && schema.nullable) {
    return {
      ...schema,
      type: [schema.type, 'null'],
    };
  }

  return schema;
};
```

This does not work with nested properties, meaning prism was failing to support nullable properly (https://github.com/stoplightio/prism/issues/486) so @XVincentX experimented with writing code to support that:

```
const transformSchema = (schema: SchemaObject) => {
  if (schema) {
    if (schema.nullable)
      return {
        ...schema,
        type: [schema.type, 'null'],
      };

    if (schema.properties) {
      Object.keys(schema.properties).forEach(prop => {
        schema.properties![prop] = transformSchema(schema.properties![prop]) as SchemaObject;
      });
      if (schema.anyOf) schema.anyOf = schema.anyOf.map(transformSchema) as SchemaObject[];
      if (schema.allOf) schema.allOf = schema.allOf.map(transformSchema) as SchemaObject[];
      if (schema.not) schema.not = transformSchema(schema.not) as SchemaObject;
      if (schema.items) schema.items = transformSchema(schema.items) as SchemaObject;
      if (schema.additionalProperties && typeof schema.additionalProperties !== 'boolean')
        schema.additionalProperties = transformSchema(schema.additionalProperties) as SchemaObject;
    }
  }

  return schema;
};
```

This is starting to look an awful lot like what mikunn/openapi-schema-to-json-schema sets out to do, but still doesn't cover as much, and has no testing, meaning the switch to use that library was a no-brainer for us. 

As for why you're seeing errors, it looks like you're passing JSON Schema into the OAS3 transformer, which sets out to transform OAS3 into JSON Schema. Can we look into why that's happening?

marbemac:
> you're looking at the wrong library

Ah, in any case tomato tomahto haha. Here's relevant same code over there: https://github.com/mikunn/openapi-schema-to-json-schema/blob/master/lib/converters/schema.js#L69-L71

Or another throw: https://github.com/mikunn/openapi-schema-to-json-schema/blob/master/lib/converters/parameter.js#L13

> As for why you're seeing errors, it looks like you're passing JSON Schema into the OAS3 transformer, which sets out to transform OAS3 into JSON Schema. Can we look into why that's happening?

I think my point might have been missed. We are not passing JSON Schema into the operation transformer. Our users are writing the specifications, and we are doing a million different things to try to help them manage the specification. You're coming at it from the point of view of a producer of openapi. We are not producing openapi, we're providing tooling to help others produce openapi that we have no control over. We need to be as fault tolerant as possible, so strict fail early libraries like this one that throw errors don't work well in our world.

For example, the library would be usable IF we updated it to either:

1. accumulate and report spec errors (but still proceed with parsing) rather than throw and halt. this overlaps slightly with what spectral should be doing.
2. add a `loose` option, that does not throw errors

---

Doing my best to explain - it's slightly nuanced in that we have to come at these problems from a fault tolerant (totally not pure openapi or json schema) point of view as a tooling vendor working with specs outside of our control. So most of our tooling has to be less "pure", e.g. expect the provided specifications to be all sorts of wrong, and do our best to continue providing value with our editors, mock servers, etc.

A good example here is that IF an error is thrown for any reason, it will block the creation of the operation object, and result in:

- prism not being to mock it at all
- docs not rendering at all
- won't show up in studio search, since no operation object
- more examples, but you get the point :)

We can be more forgiving than this, and report spec issues via another mechanism like spectral.

Happy to hop on a call or discuss further, because this concept is important across our ecosystem, not just this one case.

philsturgeon:
@marbemac Ah ok I gotcha, I seem to remember people who are _only_ using OpenAPI to make docs with ReDoc using type arrays because it doesn't shout at them. I can imagine people switching over to use Stoplight and wondering why they're getting shouted at. A loose mode, compat mode, whatever, would certainly be welcome here. 

I think we can get a PR in to that chap @mikunn I've had fairly speedy replies in the past. Let me make an issue for it. Is it ok to keep studio pegged on 1.0.6 for now until we get it resolved?

PR #15, opened 2019-08-07
Do not flatten securities

lag-of-death:
## Changes
Security array is not flattened now, so that `and` relation between security schemes can be respected (the relation in described in the links under *Specs* section below).

The issue: https://github.com/stoplightio/prism/issues/509

## Specs:
OAS3: https://swagger.io/docs/specification/authentication/#multiple
OAS2: https://swagger.io/docs/specification/2-0/authentication/#multiple


src/oas2/transformers/__tests__/securities.test.ts
```diff
@@ -95,7 +95,7 @@ describe('securities', () => {
         name: 'a name',
       };
 
-      expect(translateToSecurities([basicSecurity, implicitSecurity, apiSecurity])).toMatchSnapshot();
+      expect(translateToSecurities([[basicSecurity, implicitSecurity, apiSecurity]])).toMatchSnapshot();
     });
   });
 });
```
chris-miaskowski: Please add some tests that will verify the `OR` case (return array of more than one arrays)

src/oas2/__tests__/accessors.test.ts
```diff
@@ -87,7 +87,7 @@ describe('accessors', () => {
             },
```
chris-miaskowski: Looks good but needs some more tests: something that can prove we support both AND & OR cases.

src/oas2/transformers/securities.ts
```diff
@@ -76,6 +76,8 @@ export function translateToSingleSecurity(security: DeepPartial<Security>) {
   return;
 }
 
-export function translateToSecurities(securities: Array<DeepPartial<Security>>): HttpSecurityScheme[] {
-  return compact(securities.map(translateToSingleSecurity));
+export function translateToSecurities(securities: Security[][]): HttpSecurityScheme[][] {
+  return securities.map((security: Array<DeepPartial<Security>>) => {
```
chris-miaskowski:
```suggestion
  return securities.map((security) => {
```

Won't this be inferred?

src/oas3/accessors.ts
```diff
@@ -1,13 +1,20 @@
...
-): SecuritySchemeObject[] {
-  const globalSchemes = uniqFlatMap(spec.security);
-  const operationSchemes = uniqFlatMap(operation.security);
-  return values(pick(get(spec, 'components.securitySchemes'), operation.security ? operationSchemes : globalSchemes));
+import { get } from 'lodash';
+import { OpenAPIObject, OperationObject } from 'openapi3-ts';
+
+import { mapToKeys } from '../utils';
+
+export function getSecurities(spec: Partial<OpenAPIObject>, operation: Partial<OperationObject>): any {
```
chris-miaskowski: Did you mean?

```suggestion
export function getSecurities(spec: Partial<OpenAPIObject>, operation: Partial<OperationObject>): SecuritySchemeObject[][] {
```

Shouldn't be `any`

lag-of-death:
@chris-miaskowski, the tests are added :)

src/oas2/__tests__/accessors.test.ts
```diff
@@ -1,8 +1,10 @@
 import { Dictionary } from '@stoplight/types';
-import { Security } from 'swagger-schema-official';
+import { Security as SecurityDefinition } from 'swagger-schema-official';
```
XVincentX: I'd try *not* to rename imports unless it is really necessary.

src/oas2/__tests__/accessors.test.ts
```diff
@@ -20,13 +22,90 @@ const securityDefinitionsFixture: Dictionary<Security> = {
 };
 
 describe('accessors', () => {
-  const securityFixture = [
+  const securityFixture: Security[] = [
```
XVincentX: Maybe it's not really worth to rename the security definition on top and create a new alias just because of this fixture.

src/oas2/__tests__/accessors.test.ts
```diff
@@ -20,13 +22,90 @@ const securityDefinitionsFixture: Dictionary<Security> = {
...
     {
       api_key: [],
       petstore_auth: ['write:pets', 'read:pets'],
     },
   ];
 
+  describe('relation between schemes', () => {
+    describe('when all fo the given schemes are expected to be validated against', () => {
```
XVincentX:
```suggestion
    describe('when all of the given schemes are expected to be validated against', () => {
```

src/oas2/__tests__/accessors.test.ts
```diff
@@ -20,13 +22,90 @@ const securityDefinitionsFixture: Dictionary<Security> = {
...
+                'write:pets': 'modify pets in your account',
+              },
+              type: 'oauth2',
+            },
+          ],
+        ]);
+      });
+    });
+
+    describe('when one fo the given schemes is expected to be validated against', () => {
```
XVincentX:
```suggestion
    describe('when one of the given schemes is expected to be validated against', () => {
```

src/oas2/__tests__/accessors.test.ts
```diff
@@ -20,13 +22,90 @@ const securityDefinitionsFixture: Dictionary<Security> = {
...
+      it('returns arrays containing one element each', () => {
+        const securityFixtureWithOrRelation: Security[] = [
+          {
+            petstore_auth: ['write:pets', 'read:pets'],
+          },
+          {
+            api_key: [],
+          },
```
XVincentX: If I understand correctly this is not an OR relationship, this is *AND*

src/oas2/accessors.ts
```diff
@@ -22,21 +26,19 @@ export function getConsumes(spec: Partial<Spec>, operation: Partial<Operation>)
...
-          if (def) {
-            const defCopy = merge<Object, Security>({}, def);
-            return defCopy;
-          }
-          return null;
-        }),
-      );
-    }),
-  );
+): any {
```
XVincentX: Why `any` here?

src/oas2/transformers/securities.ts
```diff
@@ -76,6 +76,8 @@ export function translateToSingleSecurity(security: DeepPartial<Security>) {
   return;
 }
 
-export function translateToSecurities(securities: Array<DeepPartial<Security>>): HttpSecurityScheme[] {
-  return compact(securities.map(translateToSingleSecurity));
+export function translateToSecurities(securities: Security[][]): HttpSecurityScheme[][] {
+  return securities.map(security => {
```
XVincentX: You can make this one line

`return securities.map(security => compact(security.map(translateToSingleSecurity)))`

src/oas3/accessors.ts
```diff
@@ -1,13 +1,23 @@
...
-): SecuritySchemeObject[] {
-  const globalSchemes = uniqFlatMap(spec.security);
-  const operationSchemes = uniqFlatMap(operation.security);
-  return values(pick(get(spec, 'components.securitySchemes'), operation.security ? operationSchemes : globalSchemes));
+): SecuritySchemeObject[][] {
+  const globalSchemes = mapToKeys(spec.security);
+  const operationSchemes = mapToKeys(operation.security);
+
+  const opSchemesPairs = operation.security ? operationSchemes : globalSchemes;
+  const definitions = get(spec, 'components.securitySchemes');
```
XVincentX: You can provide a default value so you can avoid to check for `!definitions`
```suggestion
  const definitions = get(spec, 'components.securitySchemes', []);
```

src/oas3/accessors.ts
```diff
@@ -1,13 +1,23 @@
...
+
+  return !definitions
+    ? []
+    : opSchemesPairs.map((opSchemePair: string[]) => {
```
XVincentX: If you're only returning, there's no need to create a complete function
```ts
opSchemesPairs.map(opSchemePair => opSchemePair.map(opScheme => definitions[opScheme]));
```

src/utils.ts
```diff
@@ -1,10 +1,10 @@
-import { flatten, map, uniq } from 'lodash';
+import { map } from 'lodash';
 import * as URIJS from 'urijs';
 
 import { pathToPointer, pointerToPath } from '@stoplight/json';
 
-export function uniqFlatMap<T>(collection?: T[]) {
-  return uniq(flatten(map(collection, Object.keys)));
+export function mapToKeys<T>(collection?: T[]) {
```
XVincentX: Doesn't [mapKeys](https://lodash.com/docs/4.17.15#mapKeys) do what you need?

src/oas3/accessors.ts
lag-of-death: @XVincentX, this is not the same though. 

If I do this, then we have a problem here:
https://github.com/stoplightio/http-spec/blob/52e16cc15df040e227f06de324a951d19103c54f/src/oas3/__tests__/accessors.test.ts#L62
This test does not provide `securitySchemes`. And if I add that default value of `[]`, then we end up with the test resulting in:

```
- Snapshot
+ Received

- Array []
+ Array [
+   Array [
+     undefined,
+   ],
+ ]
```

This happens because the operation uses `specScheme` for security. If we provide `securitySchemes` as [], then we end up with `[]['specScheme']`.

src/utils.ts
lag-of-death: @XVincentX, I'm not sure really. I'm mapping an array of objects to an array of arrays of these objects' keys. Can't find usage for `_.mapKeys` here, but maybe you had something on your mind?

src/oas2/__tests__/accessors.test.ts
lag-of-death: @XVincentX, this is `OR` (https://github.com/stoplightio/http-spec/blob/40232d757aca32b5ceb0a35d152b04a5842111ff/src/oas2/__tests__/accessors.test.ts#L68). And this https://github.com/stoplightio/http-spec/blob/40232d757aca32b5ceb0a35d152b04a5842111ff/src/oas2/__tests__/accessors.test.ts#L34 is `AND`. Hope this helps!

lag-of-death:
thanks for the feedback @XVincentX! I've addressed most of it and left some comments for the rest. Please take a look :)

src/utils.ts
XVincentX: Yeah I believe you're right. In any case it's not a big deal.

src/oas3/accessors.ts
XVincentX: Aaah yeah I see what you mean. My bad. This is ok.

src/oas2/__tests__/accessors.test.ts
XVincentX: 👍

marbemac:
> I'd just hold off on merging to give @marbemac a chance to check out this as well since (again, I might be wrong) I recall he was behind the idea of flattening the security specifications.

This was just an oversight on my part. Looks good to me!

---

@lottamus could you create an issue to update elements for https://github.com/stoplightio/types/pull/50?

@pytlesk4 could you do a quick investigation of updating analyzer to the latest types package (that includes the PR linked above), and if there are any problems then perhaps create an issue to tackle it soon?

lag-of-death:
@lottamus, @pytlesk4, I'm a bit unsure here, is this PR OK to be merged? cc @marbemac

src/oas2/__tests__/accessors.test.ts
lottamus: @chris-miaskowski @lag-of-death is there anything left to address here?

src/oas2/__tests__/accessors.test.ts
```diff
@@ -20,13 +20,90 @@ const securityDefinitionsFixture: Dictionary<Security> = {
...
+            securityFixtureWithAndRelation,
+          ),
+        ).toEqual([
+          [
+            {
+              in: 'header',
+              name: 'api_key',
+              type: 'apiKey',
+            },
+            {
```
lottamus: Is there any reason you can think of that we might need the security array's key here? Currently there is nothing tying this oauth2 object back to the security array. Does that make sense?

```suggestion
            {
              securityKey: 'petstore_auth',
```

src/oas2/__tests__/accessors.test.ts
lag-of-death: @lottamus, no, the test was added:
https://github.com/stoplightio/http-spec/pull/15#discussion_r315112668

src/oas2/__tests__/accessors.test.ts
lag-of-death: @lottamus, I think that given https://github.com/stoplightio/http-spec/pull/15/files#diff-f5ac86d0dd4d4fa1390aca28ac1fbc59L107-L112 it's OK to just not add this `securityKey`? If it was not a problem before, it shouldn't be a problem now. We can address it later if it turns out to be an issue

lag-of-death:
@lottamus 
> Also I find it a bit weird that we're mapping to an array of arrays.

I find it a lot easier to work with arrays instead of objects, that's why such a structure

PR #18, opened 2019-08-26
fix: clone schema when transforming

XVincentX:
Our reference resolver is caching the references which is good — because of this though we're basically trying to re-transform a schema that has been transformed already.

Cloning the schema makes sure we're not touching the *original, cached* reference.

This is required to make sure we can work on https://github.com/stoplightio/prism/issues/552

### DIFF
--- a/src/oas3/transformers/__tests__/content.test.ts
+++ b/src/oas3/transformers/__tests__/content.test.ts
@@ -132,6 +132,27 @@ describe('translateMediaTypeObject', () => {
     };
     expect(() => testedFunction()).not.toThrow();
   });
+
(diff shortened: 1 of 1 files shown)


src/oas3/transformers/content.ts
```diff
@@ -84,7 +84,7 @@ export function translateMediaTypeObject(
 ): IMediaTypeContent {
   return {
     mediaType,
-    schema: schema ? (toJsonSchema(schema, { cloneSchema: false }) as JSONSchema4) : undefined,
+    schema: schema ? (toJsonSchema(schema, { cloneSchema: true }) as JSONSchema4) : undefined,
```
P0lip: Could add a test?

PR #19, opened 2019-08-28
Keep example property after the schema transformation

XVincentX:


src/oas3/transformers/content.ts
```diff
@@ -84,7 +84,7 @@ export function translateMediaTypeObject(
 ): IMediaTypeContent {
   return {
     mediaType,
-    schema: schema ? (toJsonSchema(schema, { cloneSchema: true }) as JSONSchema4) : undefined,
+    schema: schema ? (toJsonSchema(schema, { cloneSchema: true, keepNotSupported: ['example'] }) as JSONSchema4) : undefined,
```
philsturgeon: what about things like readOnly, writeOnly and deprecated?  They exist in OpenAPI Schema Objects but do not exist in JSON Schema draft 4. Do we have tests around this or are we stripping them out?

philsturgeon: OH I see, ok. Check default is preserved then we're ok.

PR #20, opened 2019-08-29
fix: correctly handle allowEmptyValue

XVincentX:
This is required for https://github.com/stoplightio/prism/issues/499

Correctly handle the `allowEmptyValue` transforming it in the correct JSON Schema representation.


src/oas2/transformers/__tests__/params.test.ts
```diff
@@ -165,16 +166,17 @@ describe('params.translator', () => {
               type: 'array',
             },
             int: {
+              allowEmptyValue: false,
               description: 'desc',
               maximum: 3,
               minimum: 0,
+              minLength: 1,
```
marbemac: Why move `minLength`? Pretty sure it's a string validation - the equivalent number validation is `minimum`.

XVincentX: Eh very good question. The reality is that it's not a string validation rule, but a deserialisation rule which is incorrectly handled here.

That value is fundamentally to say "can this query string parameter be empty?"

As I commented in https://github.com/stoplightio/prism/issues/499 — this is not a long term solution, as this part should probably be completely removed and handled in Prism directly; however this would (although in a monkey way) close the issue there.

I suggest to merge this and then rework on the validation side soon in Prism; but I am ok if you do not want this merged and take it seriously as a next step. @philsturgeon what's your take?

philsturgeon: I think you changed too much here. No? Marc is asking why the change from maxLength to maximum and minimum. 

The issue you mention was just suggesting:

```
if ('allowEmptyValue' in param && param.allowEmptyValue === false) {
    schema.minLength = 1;
}
```

The missing `=== false` was the problem, as it was adding `minLength: 1` when allowEmptyValue was true. We need to demand a value when empty is not allowed. 

Does this make sense? I might be wrong, but I think you could revert everything but the `=== false`.

XVincentX:
@philsturgeon That's all I did indeed :) All the other changes are because of the tests; probably the input fixture is just wrong, let me see if I can fix it.

XVincentX:
@philsturgeon You were right, some fixture were effectively wrong and invalid. I've fixed these and they should now be ok.

PR #21, opened 2019-08-30
fix: correctly set csv type to comma delimited

XVincentX:


philsturgeon:
Got some unit tests? :)

XVincentX:
No, but I do not mind writing some.

XVincentX:
@philsturgeon Although they were using snapshot, there were effectively unit tests. I've modified them to not use snapshots anymore and assert the property presence and value directly. I do not believe there's anything else we can do here.

lag-of-death:
looks good :+1: let's add some description though ;)

PR #22, opened 2019-09-02
fix(content): include whole example

P0lip:
Addresses https://github.com/stoplightio/studio/issues/80.
The question is - why did we decide to pick property `value` only? If it was intended, perhaps the issue reported in Studio is not an actual bug but a feature?


XVincentX:
I think this is a breaking change though

PR #23, opened 2019-09-07
fix: use phil's openapi schema to json schema

pytlesk4:


src/oas3/transformers/__tests__/content.test.ts
```diff
@@ -140,12 +141,11 @@ describe('translateMediaTypeObject', () => {
   });
 
   describe('schema integrity after transform', () => {
-    const schema = {
-      type: 'string',
-      nullable: true,
+    const schema = ({
+      type: ['string', 'object'],
```
marbemac: What's this change about?

marbemac:
@philsturgeon I know we want to avoid using forks, but this feels like an exception. It's affecting lots of downstream products in significant ways - e.g. if the user has something as small as a $schema property in a referenced model then docs for the endpoint do not work, publish / analyze process not working, prism not working. And we haven't had a response re when the PR will be merged. We can remove this fork when the PR is merged upstream.

philsturgeon:
@marbemac all previous replies with the author were within a day or two. It probably fell through the cracks the first time and we bumped him on Thursday. We said we'd wait a day, now it's the weekend. On Monday we can spin up a fork, but it would have been premature to do it on Friday and we're the only two working today. 😅

src/oas3/transformers/content.ts
```diff
@@ -85,7 +86,11 @@ export function translateMediaTypeObject(
   return {
     mediaType,
     schema: schema
-      ? (toJsonSchema(schema, { cloneSchema: true, keepNotSupported: ['example'] }) as JSONSchema4)
+      ? (toJsonSchema(schema, {
+          cloneSchema: true,
+          strictMode: false,
+          keepNotSupported: ['example'],
+        }) as JSONSchema4)
```
philsturgeon: Has it got a JSONSchema5 tucked away in there? They should be the same but OAS2=JS4 and OAS3=JS5 so if we can use the right interface that'd potentially remove a bit of confusion somewhere.

src/oas3/transformers/__tests__/content.test.ts
pytlesk4: Just to make sure it doesn’t throw an error.

package.json
```diff
@@ -28,13 +28,16 @@
     "test": "jest",
     "test.prod": "yarn lint && yarn test --coverage --maxWorkers=2"
   },
+  "resolutions": {
```
philsturgeon: Are we going to need to add this to downstream dependencies? We've had a bunch of issues with forks causing broken stuff because we did not add the resolution to each package.json

pytlesk4:
@philsturgeon and @marbemac why does it look spooky, referring to the scheme in the test?

package.json
pytlesk4: I think we should just bump the version every where the http spec is being used. Then we don’t run into the resolution issue.

marbemac: We shouldn't need this resolution here in general? I think it can be removed. Resolutions is kind of case by case basis depending on the package involved - assuming there are no breaking changes, generally they should be set in end products imho. In this case I think we'd need to add a resolution to studio (for studio desktop), app-gen3, and stoplight/cli.

src/oas3/transformers/__tests__/content.test.ts
philsturgeon: You replied in the main thread and also in here, I'll just reply in here. 

It looks spooky because you are changing the input from OpenAPI to JSON Schema. If the input of this test is meant to be OpenAPI and it no longer is, are we confident the code is working? 

If my change is fixing the test, we dont also need to change the inputs, right? 

Maybe you could explain why we need both my patch and this change, so it feels less concerning?

marbemac:
Oh no knock on author of openapi-schema-to-json-schema at all. Just saying that we have a means to fix this issue on our own, and we should take it in this case (fair enough though, can take it tomorrow assuming no movement on the PR by then).

src/oas3/transformers/content.ts
pytlesk4: I don’t know why that cast was added there, it was already there when I made the changes. What does that open api library actually return?

src/oas3/transformers/__tests__/content.test.ts
marbemac: > Just to make sure it doesn’t throw an error.

Suggest adding a new test for this - will clear up the confusion.

package.json
marbemac: Or what tom said, and no resolutions needed.

philsturgeon: Ok, I remember something about Prism binaries breaking and somebody else struggling to NPM behind a corp firewall for some reason, but if you think relying on resolved is ok then sure thing. Maybe @XVincentX remembers the specifics.

philsturgeon:
I didn't think anyone was knocking anyway :) Just trying to figure out if its urgent enough to do extra work on the weekend instead of giving them Monday (Finland gets a good headstart on us).

marbemac:
Yeah it won't matter until we can actually get the downstream stuff released into production anyways, so won't be before tomorrow at the earliest. So all good! Mostly tom just finding some time to get this started, and me working through emails on a Sunday, and getting to this PR review notification :).

package.json
XVincentX: In general packages coming from GitHub can be a pain because its scripts aren't executed — so if some of these have something to compile/move/anything, it won't work.

I do not think there should be a problem in this case, but in general less we do this the better.

src/oas3/transformers/__tests__/content.test.ts
pytlesk4: Done!

PR #25, opened 2019-09-20
Do not create default produces

XVincentX:
For some reason that I can't really remember OAS2 produces consumes default to `*/*`, even if the document has nothing in it.

On the other hand, in OAS3 this default mechanism is not a thing. This discrepancy is giving us troubles in Prism since we need to write special code to accomplish the same thing.

Unfortunately we lost the history because of the new repository — but if anybody recalls the reason, please speak up. Otherwise, this will uniform the behaviour between oas2 and 3.

### DIFF
--- a/src/oas2/__tests__/accessors.test.ts
+++ b/src/oas2/__tests__/accessors.test.ts
@@ -216,11 +216,6 @@ describe('accessors', () => {
   });
 
   describe('getProduces', () => {
-    test('given all empty arrays should return asterisk', () => {
-      expect(getProduces({}, {})).toEqual(['*/*']);
-      expect(getProduces({ produces: [] }, { produces: [] })).toEqual(['*/*']);
-    });
-
     test('should fallback to spec produces', () => {
       expect(getProduces({ produces: ['text/plain'] }, {})).toEqual(['text/plain']);
     });
@@ -231,11 +226,6 @@ describe('accessors', () => {
   });
 
   describe('getConsumes', () => {
-    test('given all empty arrays should return asterisk', () => {
-      expect(getConsumes({}, {})).toEqual(['*/*']);
-      expect(getConsumes({ consumes: [] }, { consumes: [] })).toEqual(['*/*']);
-    });
-
     test('should fallback to spec consumes', () => {
       expect(getConsumes({ consumes: ['text/plain'] }, {})).toEqual(['text/plain']);
     });
--- a/src/oas2/__tests__/accessors.test.ts
+++ b/src/oas2/__tests__/accessors.test.ts
@@ -216,6 +216,11 @@ describe('accessors', () => {
   });
 
   describe('getProduces', () => {
+    test('given all empty arrays should return an empty array', () => {
+      expect(getProduces({}, {})).toEqual([]);
+      expect(getProduces({ produces: [] }, { produces: [] })).toEqual([]);
+    });
+
     test('should fallback to spec produces', () => {
       expect(getProduces({ produces: ['text/plain'] }, {})).toEqual(['text/plain']);
     });
@@ -226,6 +231,11 @@ describe('accessors', () => {
   });
 
   describe('getConsumes', () => {
+    test('given all empty arrays should return an empty array', () => {
+      expect(getConsumes({}, {})).toEqual([]);
+      expect(getConsumes({ consumes: [] }, { consumes: [] })).toEqual([]);
+    });
+
     test('should fallback to spec consumes', () => {
       expect(getConsumes({ consumes: ['text/plain'] }, {})).toEqual(['text/plain']);
     });
--- a/src/oas2/accessors.ts
+++ b/src/oas2/accessors.ts
@@ -43,5 +43,5 @@ function getSecurity(
 
 function getProducesOrConsumes(which: 'produces' | 'consumes', spec: Partial<Spec>, operation: Partial<Operation>) {
   const mimeTypes = get(operation, which, get(spec, which, [])) as string[];
-  return mimeTypes.length ? mimeTypes : ['*/*'];
+  return mimeTypes;
 }
--- a/src/oas2/operation.ts
+++ b/src/oas2/operation.ts
@@ -1,6 +1,6 @@
-import { DeepPartial, IHttpOperation } from '@stoplight/types';
+import { IHttpOperation } from '@stoplight/types';
 import { get, isNil, omitBy } from 'lodash';
-import { Operation, Parameter, Path, Response, Spec } from 'swagger-schema-official';
+import { Operation, Parameter, Path, Response } from 'swagger-schema-official';
 
 import { getOasParameters } from '../oas/accessors';
 import { translateToTags } from '../oas/tag';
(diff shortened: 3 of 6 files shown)


src/oas2/accessors.ts
```diff
@@ -43,5 +43,5 @@ function getSecurity(
 
 function getProducesOrConsumes(which: 'produces' | 'consumes', spec: Partial<Spec>, operation: Partial<Operation>) {
   const mimeTypes = get(operation, which, get(spec, which, [])) as string[];
-  return mimeTypes.length ? mimeTypes : ['*/*'];
```
lag-of-death: @XVincentX, why do we need this? Is this consistent with OAS3?

src/oas2/__tests__/accessors.test.ts
```diff
@@ -216,11 +216,6 @@ describe('accessors', () => {
   });
 
   describe('getProduces', () => {
-    test('given all empty arrays should return asterisk', () => {
-      expect(getProduces({}, {})).toEqual(['*/*']);
```
lag-of-death: would
```
expect(getProduces({}, {})).toEqual([])
```
be OK?

src/oas2/operation.ts
```diff
@@ -11,11 +11,6 @@ import { translateToResponses } from './transformers/responses';
 import { translateToSecurities } from './transformers/securities';
 import { translateToServers } from './transformers/servers';
 
-// TODO
-export function computeOas2Operations(spec: DeepPartial<Spec>): IHttpOperation[] {
-  return [];
```
lag-of-death: yeah, I thought this was dead code :+1:

src/oas2/accessors.ts
XVincentX: @lag-of-death This is the inconsistency you found while working on the PR yesterday; empty content in OAS3, but */* for OAS2 — which is technically incorrect. Does that clarify?

lag-of-death: @XVincentX, I'm sorry, I must have been colorblind there for a sec.. all good, my bad!

lottamus:
Is this a breaking change? Not sure if that default is being expected anywhere else in our ecosystem.

Here's the PR where you changed the default mimetype from `["*"]` to `["*/*"]` if this helps any: https://github.com/stoplightio/graphite/pull/191

> This PR will make sure that the media type produced for default requests/responses will be `*/*` instead of `*` — this will facilitate the matching jobs in Prism (and probably some other software as well)

XVincentX:
@lottamus Thanks for the check!

This PR is effectively renaming the default content (which is put when no produces/consumes is found) — but it does not explain why we put it.

I think we should either:

1. Remove this from OAS2 adapter
2. Add this in the OAS3 adapter as well.

I am not too opinionated on this and to be honest I do not know what would be the repercussions on Studio, this is why I have involved almost everybody in this PR.

src/oas2/__tests__/accessors.test.ts
pytlesk4: @lag-of-death yes, that should be what the test is. The same thing for consumes as well.

@XVincentX What happens if there is a global consumes/produces with json in it, and then an operation has an empty array for consumes/produces?

marbemac:
Nothing on the Studio side - I believe most of this was originally driven by Prism requirements? If no consumes/produces is set in the spec or operation, but a 200 json response is defined on the operation, what will prism do right now? What will it do if we remove that `*/*`?

pytlesk4:
> Nothing on the Studio side - I believe most of this was originally driven by Prism requirements? If no consumes/produces is set in the spec or operation, but a 200 json response is defined on the operation, what will prism do right now? What will it do if we remove that `*/*`?

If there is no global produces/consumes then the response shouldn't have a content type.

XVincentX:
So everybody agrees with this change? Am I assuming correctly?

src/oas2/__tests__/accessors.test.ts
XVincentX: @pytlesk4 Basically the local overrides the global, so you would still get the local content.

pytlesk4:
```suggestion
    test('given all empty arrays should return an empty array', () => {
      expect(getProduces({}, {})).toEqual([]);
      expect(getProduces({ produces: [] }, { produces: [] })).toEqual([]);
    });
```

Just add this test back, but test for an empty array.

src/oas2/__tests__/accessors.test.ts
```diff
@@ -231,11 +226,6 @@ describe('accessors', () => {
   });
 
   describe('getConsumes', () => {
-    test('given all empty arrays should return asterisk', () => {
```
pytlesk4:
```suggestion
    test('given all empty arrays should return an empty array', () => {
      expect(getConsumes({}, {})).toEqual([]);
      expect(getConsumes({ consumes: [] }, { consumes: [] })).toEqual([]);
    });
```

XVincentX:
@marbemac @philsturgeon @pytlesk4 @lottamus 

Ok with this change merged there's one thing that you should be aware of (and agree if it's ok or not).

Fundamentally for this document:

```yaml
swagger: '2.0'
 paths:
   "/pet/findByStatus":
     post:
       produces:
       - application/json
       parameters:
       - in: body
         name: body
         schema:
           type: object
           properties:
             id:
               type: integer
               format: int64
       responses:
         '200':
           schema:
             type: string
```

Since there's nor a local nor a global consumes (making the document effectively _invalid_) — we won't have any body in the `request` node; because they're being created starting from the global or local `consumes` array.

I think this is the correct behaviour (and we can probably easily write a Spectral rule for this, I've already created an issue for it), but I want to make sure everybody is ok with it. Otherwise, I'll modify the code to accommodate this.

lottamus:
@XVincentX are global/local consumes really required to have a request body? Is that according to the spec or a standard we are trying to enforce? Is this why we were defaulting to `*/*` before?

XVincentX:
Without a local or global produces/consumes, we practically have no idea how to deserialise the payload coming in and serialise the response going out.

I've looked at the spec multiple times including the issues and apparently one of the two must be present;

if global consumes/produces are not provided (apparently they're optional), then it must be provided per each operation.

To be precise though, I've never found a precise statement about it.

lottamus:
Right, so you are you asking if we should ignore the request body and proceed as if it didn't exist, OR tell the user their spec is invalid, OR assume some default? I think you're good with doing the first two: ignoring and letting the user know why (if that's possible).

Here's [the response](https://github.com/OAI/OpenAPI-Specification/issues/1031#issuecomment-295132120) to that comment from Ron Ratovsky at SmartBear:

> The petstore sample has a few gotchas in it, but regardless, in swagger-ui we send application/json by default if no consumes is provided but there's a body payload. This is to try and help users instead of chocking on things like that. It is not, however, the default in the spec.

XVincentX:
@lottamus Ok, my only concern is that http spec won't produce the node for the body *at all*; is that ok of analyser/studio?

lottamus:
Yeah it will be confusing to the end user when the Forms Panel shows a request body but the Read Panel doesn't 🤔. But a Spectral rule would at least surface the error whether they notice it or not.

XVincentX:
I never saw such document to be honest; it would be interesting to run a scan in our database but I'd say it's not worth the effort. Let's keep it in this way and add it back if required.

PR #27, opened 2019-10-15
fix(oas3): getSecurities filters out undefined values

P0lip:
Fixes https://github.com/stoplightio/studio-internal/issues/1023


src/oas3/accessors.ts
```diff
@@ -1,17 +1,16 @@
 import { get } from 'lodash';
 import { OpenAPIObject, OperationObject, SecuritySchemeObject } from 'openapi3-ts';
 
-import { mapToKeys } from '../utils';
+import { isObject, mapToKeys } from '../utils';
```
marbemac: Would any of the following work?

- https://lodash.com/docs/4.17.14#isObject
- https://lodash.com/docs/4.17.14#isObjectLike
- https://lodash.com/docs/4.17.14#isPlainObject

src/oas3/accessors.ts
```diff
@@ -1,17 +1,16 @@
...
 
 export function getSecurities(
   spec: Partial<OpenAPIObject>,
```
marbemac: Should we change to `DeepPartial` here and in the operation property?

In general using deep partial on entire open api specifications, and then handling all the property checks manually everywhere, is going to succkk. Wondering if there is a better way? Lodash? Although I suppose we lose some typings, but in reality the underlying spec (openapi) is really not changing very often.

src/oas3/accessors.ts
P0lip: Oh, I thought `isObject` wasn't a type-guard, but it is!
Going to use it.

src/oas3/accessors.ts
P0lip: > Lodash? Although I suppose we lose some typings, but in reality the underlying spec (openapi) is really not changing very often.

Frankly, let's wait for TS 3.7 to be released so that we could utilize optional chaining operator and null null coalescing operator.
Doing so would still let us have fairly clean code, while still preserving typings.

P0lip: Seems like 3.7 beta is [out](https://devblogs.microsoft.com/typescript/announcing-typescript-3-7-beta/) already.
How do you feel about trying it out in http-spec?

PR #28, opened 2019-10-15
feat: add transformOperations

marbemac:
Turns open api file into array of http operations.


src/oas/computeOperations.ts
```diff
@@ -0,0 +1,29 @@
+import { IHttpOperation } from '@stoplight/types';
+import { flatten, get, keys, map } from 'lodash';
+import { OpenAPIObject } from 'openapi3-ts';
+import { Spec } from 'swagger-schema-official';
+
+import { HttpOperationTransformer } from '../oas/types';
+
+const methods = ['get', 'post', 'put', 'delete', 'options', 'head', 'patch', 'trace'];
+
+export function computeOasOperations(
```
pytlesk4: Nice, I like that it doesn't care that it's a oas2/3 document. Can we do something similar for the `transformOas2Operation` and `transformOas3Operation` and make a `trasnformOasOperation`?

pytlesk4: Is it required to pass in the transformers?

Can we just check the openapi or swagger property?

src/oas/computeOperations.ts
```diff
@@ -0,0 +1,29 @@
...
+  document: Spec | OpenAPIObject,
+  transformer: HttpOperationTransformer<any>,
```
P0lip:
```suggestion
  transformer: HttpOperationTransformer<unknown>,
```

pytlesk4:
@marbemac I need this function, do you have time to finish this out?

XVincentX:
> I suspect it all comes down to the fact we don't trust the input, but if we are unsure whether a proper document is provided, we could simply specify unknown and use type guards instead.

I agree with this so much. We're really, *really* indisciplined with the usage of `any` and this brought more than 1 problem in Prism, for example.

I have an idea how to solve this problem *FOREVER*, but you'd probably kill me for that :trollface:

PR #31, opened 2019-10-18
feat: oas3 callbacks support

StefanDywersant:
Introduces a capability for reading callbacks from OASv3 specs. Needs https://github.com/stoplightio/types/pull/54 to be merged first (types).

The motivation behind extending the IHttpOperation interface: callbacks in OASv3 spec are wrapped in an object with callback name as property and operation as value. SL's way of representing such objects is to convert it to an array and add a property inside with the key name (see operations array, examples array, etc.). This change is compatible with that approach.

**Master PR**: https://github.com/stoplightio/prism/pull/716
**Related issue**: https://github.com/stoplightio/prism/issues/331
**The spec**: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#callbackObject
**Callback example**: https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/callback-example.yaml

### DIFF
--- a/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
+++ b/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
@@ -156,22 +156,7 @@ Object {
   "path": "/users/{userId}",
   "request": Object {
     "body": Object {
-      "contents": Array [
-        Object {
-          "encodings": Array [],
-          "examples": Array [],
-          "mediaType": "application/json",
-          "schema": Object {
-            "$schema": "http://json-schema.org/draft-04/schema#",
-            "properties": Object {
-              "id": "string",
-            },
-            "type": "object",
-          },
-        },
-      ],
-      "description": undefined,
-      "required": undefined,
+      "contents": Array [],
     },
     "cookie": Array [],
     "headers": Array [],
--- a/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
+++ b/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
@@ -1,5 +1,64 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`transformOas3Operation callback 1`] = `
+Object {
+  "callbacks": Array [
+    Object {
+      "callbackName": "myCallback",
+      "id": "?http-operation-id?",
+      "iid": "cbId",
+      "method": "post",
+      "path": "http://prism.com?transactionId={$request.body#/id}",
+      "request": Object {
+        "body": Object {
+          "contents": Array [],
+        },
+        "cookie": Array [],
+        "headers": Array [],
+        "path": Array [],
+        "query": Array [],
+      },
+      "responses": Array [],
+      "security": Array [],
+      "servers": Array [],
+      "tags": Array [],
+    },
+  ],
+  "id": "?http-operation-id?",
+  "iid": "opid",
+  "method": "post",
+  "path": "/subscribe",
+  "request": Object {
+    "body": Object {
+      "contents": Array [
+        Object {
+          "encodings": Array [],
+          "examples": Array [],
+          "mediaType": "application/json",
+          "schema": Object {
+            "$schema": "http://json-schema.org/draft-04/schema#",
+            "properties": Object {
+              "id": "string",
+            },
+            "type": "object",
+          },
+        },
+      ],
+      "description": undefined,
+      "required": undefined,
+    },
+    "cookie": Array [],
+    "headers": Array [],
+    "path": Array [],
+    "query": Array [],
+  },
+  "responses": Array [],
+  "security": Array [],
+  "servers": Array [],
+  "tags": Array [],
+}
+`;
+
 exports[`transformOas3Operation given no tags should translate operation with empty tags array 1`] = `
 Object {
   "deprecated": true,
--- a/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
+++ b/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
@@ -38,11 +38,10 @@ Object {
           "schema": Object {
             "$schema": "http://json-schema.org/draft-04/schema#",
             "properties": Object {
-              "id": "string",
+              "id": Object {
+                "type": "string",
+              },
             },
-            "required": Array [
-              "id",
-            ],
             "type": "object",
           },
         },
--- a/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
+++ b/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
@@ -8,7 +8,7 @@ Object {
       "id": "?http-operation-id?",
       "iid": "cbId",
       "method": "post",
-      "path": "http://prism.com?transactionId={$request.body#/id}",
+      "path": "http://example.com?transactionId={$request.body#/id}",
       "request": Object {
         "body": Object {
           "contents": Array [],
--- a/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
+++ b/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
@@ -40,6 +40,9 @@ Object {
             "properties": Object {
               "id": "string",
             },
+            "required": Array [
+              "id",
+            ],
             "type": "object",
           },
         },
--- a/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
+++ b/src/oas3/__tests__/__snapshots__/operation.test.ts.snap
@@ -156,7 +156,22 @@ Object {
   "path": "/users/{userId}",
   "request": Object {
     "body": Object {
-      "contents": Array [],
+      "contents": Array [
+        Object {
+          "encodings": Array [],
+          "examples": Array [],
+          "mediaType": "application/json",
+          "schema": Object {
+            "$schema": "http://json-schema.org/draft-04/schema#",
+            "properties": Object {
+              "id": "string",
+            },
+            "type": "object",
+          },
+        },
+      ],
+      "description": undefined,
+      "required": undefined,
     },
     "cookie": Array [],
     "headers": Array [],
--- a/src/oas3/operation.ts
+++ b/src/oas3/operation.ts
@@ -42,7 +42,7 @@ export const transformOas3Operation: Oas3HttpOperationTransformer = ({ document,
       getOasParameters(operation.parameters as ParameterObject[], pathObj.parameters),
       operation.requestBody as RequestBodyObject,
     ),
-    callbacks: translateToCallbacks(operation.callbacks),
+    callbacks: translateToCallbacks(operation.callbacks || {}),
     tags: translateToTags(operation.tags || []),
     security: translateToSecurities(getSecurities(document, operation)),
   };
--- a/src/oas3/operation.ts
+++ b/src/oas3/operation.ts
@@ -6,6 +6,7 @@ import { getOasParameters } from '../oas/accessors';
 import { translateToTags } from '../oas/tag';
 import { Oas3HttpOperationTransformer } from '../oas/types';
 import { getSecurities } from './accessors';
+import { translateToCallbacks } from './transformers/callbacks';
 import { translateToRequest } from './transformers/request';
 import { translateToResponses } from './transformers/responses';
 import { translateToSecurities } from './transformers/securities';
@@ -41,6 +42,7 @@ export const transformOas3Operation: Oas3HttpOperationTransformer = ({ document,
       getOasParameters(operation.parameters as ParameterObject[], pathObj.parameters),
       operation.requestBody as RequestBodyObject,
     ),
+    callbacks: translateToCallbacks(operation.callbacks),
     tags: translateToTags(operation.tags || []),
     security: translateToSecurities(getSecurities(document, operation)),
   };
--- a/src/oas3/operation.ts
+++ b/src/oas3/operation.ts
@@ -42,7 +42,7 @@ export const transformOas3Operation: Oas3HttpOperationTransformer = ({ document,
       getOasParameters(operation.parameters as ParameterObject[], pathObj.parameters),
       operation.requestBody as RequestBodyObject,
     ),
-    callbacks: translateToCallbacks(operation.callbacks || {}),
+    callbacks: operation.callbacks && translateToCallbacks(operation.callbacks),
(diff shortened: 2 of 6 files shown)


src/oas3/transformers/callbacks.ts
```diff
@@ -0,0 +1,28 @@
+import { IHttpCallbackOperation } from '@stoplight/types';
+import { CallbacksObject } from 'openapi3-ts';
+import { transformOas3Operation } from '../operation';
+
+export function translateToCallbacks(callbacks?: CallbacksObject): IHttpCallbackOperation[] | undefined {
```
pytlesk4: Isn't a callback just an HTTP operation basically? Why create a new type?

Just a question, I know this is a draft, but glad we are thinking about supporting it.

StefanDywersant: Yeah, callback is an operation that is wrapped in an object with callback names as keys (see `myWebhook` here https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#callback-object-example).

My idea is to extend HttpOperation with callback name to achieve two goals:
- keep callbacks in array
- keep callback names for further use

Type is defined here: https://github.com/stoplightio/types/pull/54/files#diff-6b2dc2c69dd576eb785765cc8d3bb904R44

Btw. I'm working on webhooks https://github.com/stoplightio/prism/issues/331

src/oas3/transformers/callbacks.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!callbacks) return;
+
+  return Object.entries(callbacks).reduce((results: IHttpCallbackOperation[], [callbackName, path2Methods]) => {
+    for (const [path, method2Op] of Object.entries(path2Methods)) {
+      for (const [method, op] of Object.entries(method2Op as { [key: string]: {} })) {
+        results.push({
+          ...transformOas3Operation({
```
marbemac: 🤔 kind of odd recursive thing going on here. Probably need to give it the complete original document instead of this shim - `transformOas3Operation` uses data like `operation.servers || pathObj.servers || document.servers` and the security schemes from the oas document, which is not included here.

StefanDywersant: Right, this may look odd. However:

- I don't think servers make sense inside callback operation. CB op calls different server than those defined in top-level servers prop. The called server is defined inside CB path. That's why I think it is reasonable to leave servers empty in CB op.

- Security schemes also make little sense for callback op. OAS spec defines security schemes of the API itself. Callback is a bit different. It targets some different host and most of the time security is implemented via apiKey in url. See the example below:
```
CLIENT                   API
call /subscribe
                         call {$response.body#/url}&apiKey={$response.body#/apiKey}
```

So if we agree that security and servers are not applicable for callback operation let's think whether `IHttpCallbackOperation` should contain `servers` and `security` fields at all. On one hand, it makes it fully compatible with `IHttpOperation`, on the other, it makes no sense in context of the callback.

WDYT about creating IHttpCallbackOperation based on picking/omitting fields from IHttpOperation. Instead of fully inheriting (as it is now).

StefanDywersant: @philsturgeon @XVincentX curious your opinion too!

src/oas3/transformers/callbacks.ts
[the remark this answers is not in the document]
StefanDywersant: I changed the input arg to required. I vote on keeping function return value nullable. The reason is that it is most convenient way of not adding callbacks field to all `IHttpOperation` instances. I don't want to check for `Object.entries(operation).length` at `transformOas3Operation` level because it won look nice.

src/oas3/transformers/callbacks.ts
marbemac: > It targets some different host and most of the time security is implemented via apiKey in url

Most of the time, or all of the time? ;)

Is token auth in a header not a thing? Or in general, how would one indicate that a callback requires apikey auth in the url? Pretty sure that callbacks can indicate that they use xyz security scheme(s) defined in the root oas document, similar to regular http operations, so we probably need to support that - what do you think?

src/oas3/__tests__/__snapshots__/operation.test.ts.snap
```diff
@@ -1,5 +1,64 @@
...
 
+exports[`transformOas3Operation callback 1`] = `
+Object {
+  "callbacks": Array [
+    Object {
+      "callbackName": "myCallback",
+      "id": "?http-operation-id?",
+      "iid": "cbId",
+      "method": "post",
+      "path": "http://prism.com?transactionId={$request.body#/id}",
```
philsturgeon: Let's not use domains we dont own. if its not localhost or example.org we could have weird things happening accidentally talking to other peoples servers.

src/oas3/transformers/callbacks.ts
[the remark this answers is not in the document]
XVincentX: Let's return `Option< IHttpCallbackOperation[]>` :trollface: 

Oh no, this is not Prism.

src/oas3/operation.ts
```diff
@@ -41,6 +42,7 @@ export const transformOas3Operation: Oas3HttpOperationTransformer = ({ document,
       getOasParameters(operation.parameters as ParameterObject[], pathObj.parameters),
       operation.requestBody as RequestBodyObject,
     ),
+    callbacks: translateToCallbacks(operation.callbacks || {}),
```
XVincentX:
```suggestion
    callbacks: operations.callbacks ? translateToCallbacks(operation.callbacks) : undefined,
```

I know that this code will look different than the other one, but we seriously need to stop this habit that everything can be undefined and somebody else will figure it out.

src/oas3/transformers/callbacks.ts
XVincentX: >  Probably need to give it the complete original document 

I have checked out the code of that function and in reality only 1% of the document is used and what's needed can be easily be passed as a standalone parameter instead of using the whole document. This would simplify the things for the callbacks a lot. `path, servers, security` — that's all we need. I suggest we pursue this path @StefanDywersant 

> I don't think servers make sense inside callback operation. CB op calls different server than those defined in top-level servers prop. The called server is defined inside CB path. That's why I think it is reasonable to leave servers empty in CB op.

I am on the opposite side. The spec says that the element of a Callback Object is a Path, and we should fully support it; that means we should fallback to the global servers if required. I think it's kind of regular to see callbacks to external APIs that are still under the same domain.

> Security schemes also make little sense for callback op.

Same as above, if the callback element is a Path, we should fully support it.

> I do not think I have ever seen anyone sending an API Key over the URL.

Unfortunately there are companies doing it. [Uala](https://uala.it) is one of these and I am afraid we in Stoplight are doing the same in api-gen3. We're likely leaking our apiKeys in all the logging system we're using.

marbemac:
> I have checked out the code of that function and in reality only 1% of the document is used and what's needed can be easily be passed as a standalone parameter instead of using the whole document. This would simplify the things for the callbacks a lot. path, servers, security — that's all we need. I suggest we pursue this path

Agree!

> Unfortunately there are companies doing it. Uala is one of these and I am afraid we in Stoplight are doing the same in api-gen3. We're likely leaking our apiKeys in all the logging system we're using.

Token in query string is actually relatively common. It's very convenient because can link to it and open in browser. Random example - click on the "raw" button for a private repo file in Github, and note the token on the URL https://raw.githubusercontent.com/stoplightio/studio-internal/develop/README.md?token=xxx. Side note - we support passing token in URL in our new API, but redact the query param in the logs.

src/oas3/transformers/callbacks.ts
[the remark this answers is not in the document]
StefanDywersant: `Option<NonEmptyArray<IHttpCallbackOperation[]>>` 🚎

src/oas3/__tests__/__snapshots__/operation.test.ts.snap
StefanDywersant: `http-spec` has no logic for making any requests but you right, let's keep it a general rule and use only safe domains. Changed 👍

src/oas3/transformers/callbacks.ts
philsturgeon: @XVincentX we should stop that, but also I have never seen it _for callbacks_. Sending an "API Key" in a URL would be a little wonky, most webhooks I've seen send a signature - which cannot be described in OpenAPI currently) or a HTTP Basic / Digest.

StefanDywersant: > @XVincentX we should stop that, but also I have never seen it _for callbacks_. Sending an "API Key" in a URL would be a little wonky, most webhooks I've seen send a signature - which cannot be described in OpenAPI currently) or a HTTP Basic / Digest.

Callbacks give you full flexibility in how you implement security. ApiKey is an example. I believe most secure way is to generate single-use key for each callback at the moment of calling the enclosing operation. That's it, simple and secure (see here https://sendgrid.com/blog/whats-webhook/). I think Github or Stripe implemented that signing in order make folks life easier. Both, GH and Stripe treat their signing as optional, leaving security decisions to the caller.

XVincentX: > I believe most secure way is to generate single-use key for each callback at the moment of calling the enclosing operation. That's it, simple and secure

Totally agree with you — this was never about what's the best practice, but more about "it's a scenario and even if we do not like it from a security standpoint, we might need to support it"

To complete the information, another kind of spread way is simply to ping back the original server that sent the web hook to make sure it is effectively coming from them. Paypal's been using this [for long time](https://developer.paypal.com/docs/api/webhooks/v1/#verify-webhook-signature)

> CB op calls different server

This is a strong assumption I wouldn't be so sure about.

---

I still kind of think we should fully support it; I take the argument about the security, but `servers` to me looks a legit use case. Also, I take the argument that this is a pain in the ass to support :)

We can probably go baby steps here and support the whole toy if people will ask it; so I propose

1. Ignore security and servers at the moment — we'll document the current limitation in Prism when the feature gets done
2. Try to see if we can still refactor this function to take only the parameters it really needs (path, security, servers) from the original document?

@marbemac @philsturgeon @StefanDywersant

marbemac: I'm good with the baby steps proposal!

philsturgeon:
Short lived user specific tokens are common. Long lived “API Keys” (which are usually generic for an entire application and not user specific” are incredibly unsafe, especially in the URL. 

For callbacks we should support it because it’s in the spec, but I started this sidebar conversation only to point out we shouldn’t consider it so common that it’s the only thing we support. We also shouldn’t document it as an example in case somebody decides to do it based on that example.

src/oas3/transformers/callbacks.ts
XVincentX: @StefanDywersant Pinging you on this just to make sure you do not lose the comment.

StefanDywersant: Changing the signature might be problematic, `transformOas3Operation` is directly used in `prism`, `platform` and `graphite`.

A version compatible with both would look like: `transformOas3Operation({ path, method, document, operation, pathObj, globalServers, globalSecurities})`. As you can see it is super confusing of how to use it. TS has no deprecated keyword also which would help here.

I don't like the above, I don't have name proposal for alternative version of that func. I don't think it will be handy for end-users to extract pathObj/globalSecurity/globalServer on their own.

I vote no refactor 👎

StefanDywersant: The only use case I see for `servers` is that the service defines a callback that will hit itself. I can't think about any real-world example for that. If it hits itself then why it does not do that internally? It would look like a bad design of that service.

I still consider callbacks as a `separate context` when considering `servers` and `security`.

Maybe this is a good question for OpenAPI Specification WG guys?

XVincentX: > The only use case I see for servers is that the service defines a callback that will hit itself. 

I feel you're narrowing down the domain too much. There are a lot of companies with different teams with different APIs on different files on the same server. Think simply something like

`api.stoplight.io/prism`
`api.stoplight.io/platform`
`api.stoplight.io/next`
`api.stoplight.io/studio`

It's really likely they won't be on the same document and owned by the same teams and still, callbacks within the same organisation will be a thing.

Anyway — this is becoming a product decision which I do not own. @philsturgeon @marbemac this is your call; I am ok leaving it as it is right now (no server and no security support for callbacks at the moment) but if you have another idea, please say it now.

About the refactor, I would prefer to see this done because in the current state (to me) it does not look great but I understand the frustration of getting it done. I'm ok leaving as it is right now and rework this on the next occasion.

StefanDywersant: > It's really likely they won't be on the same document and owned by the same teams and still, callbacks within the same organisation will be a thing.

I don't think there is an use-case for public-facing api. However, you convinced me that there is a use-case for documenting **internal organization APIs**.

Because of that I'd say, let's implement security and servers. 

There is quite a lot of effort to do that at prism level: faking security, constructing server (variables!). Http-spec will also require some love.

Currently, Callbacks PR is 1k LoC. Are you ok with splitting it into two PRs? (1. what we have now 2. servers and security).

> About the refactor, I would prefer to see this done because in the current state (to me) it does not look great but I understand the frustration of getting it done. I'm ok leaving as it is right now and rework this on the next occasion.

I don't see a solution that will improve the code here. E.g. your proposed solution makes that function problematic to use outside. Those lines https://github.com/stoplightio/http-spec/blob/master/src/oas3/operation.ts#L20-L28 would need to be moved to the caller space. So since the function is used by `prism`, `platform` and `graphite/studio` we would need to move those checks there.

XVincentX: No rush @StefanDywersant — this is a product decision that I can't take alone; @philsturgeon is probably the folk deciding if that's critical or not for the feature to be released.

marbemac: I think we go baby steps for now, and do security and servers later once we have real world use cases to help drive the decision.

philsturgeon: Yeah I think baby steps (not dealing with servers or security) will be fine. People always let us know what they want. 

Right now I am unsure how servers would fit in, and security is going to be a lot of work to get right, as we currently only handle validating correctness, we do not handle generating an example request with reasonable default tokens. 

Seeing as a lot of people use signatures (which OpenAPI does not directly support) we can probably just skip it entirely.

PR #32, opened 2019-10-22
fix: handle nullish headers

P0lip:
Fixes https://github.com/stoplightio/studio-internal/issues/1053

### DIFF
--- a/src/oas3/guards.ts
+++ b/src/oas3/guards.ts
@@ -1,7 +1,6 @@
-import { Dictionary } from '@stoplight/types/dist';
+import { Dictionary } from '@stoplight/types';
 import { isObject } from 'lodash';
-import { HeaderObject, SecuritySchemeObject } from 'openapi3-ts';
-import { BaseParameterObject } from 'openapi3-ts/src/model/OpenApi';
+import { BaseParameterObject, HeaderObject, SecuritySchemeObject } from 'openapi3-ts';
 
 export const isSecurityScheme = (maybeSecurityScheme: unknown): maybeSecurityScheme is SecuritySchemeObject =>
   isObject(maybeSecurityScheme) &&
--- a/src/oas3/guards.ts
+++ b/src/oas3/guards.ts
@@ -9,6 +9,8 @@ export const isSecurityScheme = (maybeSecurityScheme: unknown): maybeSecuritySch
 
 export const isBaseParameterObject = (maybeBaseParameterObject: unknown): maybeBaseParameterObject is BaseParameterObject =>
   isObject(maybeBaseParameterObject) && (
+    'description' in maybeBaseParameterObject ||
+    'required' in maybeBaseParameterObject ||
     'content' in maybeBaseParameterObject ||
     'style' in maybeBaseParameterObject ||
     'examples' in maybeBaseParameterObject ||
--- a/src/oas3/guards.ts
+++ b/src/oas3/guards.ts
@@ -1,7 +1,20 @@
 import { Dictionary } from '@stoplight/types/dist';
 import { isObject } from 'lodash';
-import { SecuritySchemeObject } from 'openapi3-ts';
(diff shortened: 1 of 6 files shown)


src/oas3/guards.ts
```diff
@@ -1,7 +1,20 @@
...
+import { BaseParameterObject } from 'openapi3-ts/src/model/OpenApi';
 
 export const isSecurityScheme = (maybeSecurityScheme: unknown): maybeSecurityScheme is SecuritySchemeObject =>
   isObject(maybeSecurityScheme) &&
   typeof (maybeSecurityScheme as Dictionary<unknown>).type === 'string';
+
+
+export const isBaseParameterObject = (maybeBaseParameterObject: unknown): maybeBaseParameterObject is BaseParameterObject =>
+  isObject(maybeBaseParameterObject) && (
+    'content' in maybeBaseParameterObject ||
```
P0lip: Too strict? `BaseParameterObject` can contain quite a few properties.

src/oas3/guards.ts
```diff
@@ -1,7 +1,20 @@
 import { Dictionary } from '@stoplight/types/dist';
 import { isObject } from 'lodash';
-import { SecuritySchemeObject } from 'openapi3-ts';
+import { HeaderObject, SecuritySchemeObject } from 'openapi3-ts';
+import { BaseParameterObject } from 'openapi3-ts/src/model/OpenApi';
```
marbemac: Don't like importing from `src` directory, and seems unimportant in usage below?

src/oas3/guards.ts
marbemac: Naw, looks good. Assuming at least one of the properties here is required to be in the parameter object.

src/oas3/guards.ts
P0lip: Ooops, good catch. WebStorm does it quite often when I perform automatic import insertion.

src/oas3/guards.ts
P0lip: The problem is that *all* properties are optional, and there is no required one.

marbemac: Looks like `name` and `in` are required. https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-locations

Should we also check for `$ref` btw? Since parameters can be $refs.

P0lip: Yeah, `in` is required, but not in `BaseParameterObject` :/
```ts
export interface ISpecificationExtension {
    // Cannot constraint to "^x-" but can filter them later to access to them
    [extensionName: string]: any;
}

export interface BaseParameterObject extends ISpecificationExtension {
    description?: string;
    required?: boolean;
    deprecated?: boolean;
    allowEmptyValue?: boolean;

    style?: ParameterStyle; // "matrix" | "label" | "form" | "simple" | "spaceDelimited" | "pipeDelimited" | "deepObject";
    explode?: boolean;
    allowReserved?: boolean;
    schema?: SchemaObject | ReferenceObject;
    examples?: { [param: string]: ExampleObject | ReferenceObject };
    example?: any;
    content?: ContentObject;
}
```
This is how it looks like. Pretty generic.

P0lip: @marbemac so, what's the final call? Shall we merge it or perhaps look for every single possible property?

marbemac: Hmm, isn't `isHeaderObject` check simply `in === header`? Slightly confused at how you're using this base parameter object thing here. Seems like what we really want to know is if it's a header object?

P0lip:
```
export interface HeaderObject extends BaseParameterObject {
}
```
This is how `HeaderObject `looks like. It's a mirror of `BaseParameterObject` :confused:

marbemac: Alright, maybe let's add `description` and `required`, and then merge it.

PR #37, opened 2019-11-05
fix: upgrade packages

XVincentX:
This PR is no really required but I messed up the previous publish process so we need something else to publish.


XVincentX:
Aha, I'll take the argument (and the punishment)

PR #39, opened 2019-11-11
fix(oas3): filter out invalid servers and server variables

P0lip:
Should help with https://github.com/stoplightio/studio-internal/issues/1212


src/oas3/transformers/servers.ts
```diff
@@ -1,19 +1,19 @@
...
-        ({
-          default: String(value.default),
-          description: String(value.default),
-          enum: map(value.enum, String),
-        } as INodeVariable),
+    variables: mapValues<Dictionary<ServerVariableObject>, INodeVariable>(
+      pickBy(server.variables, isServerVariableObject),
+      value => ({
+        default: String(value.default),
+        description: String(value.default),
```
marbemac: Make sure to get the fix from the PR that Chris just merged!

PR #40, opened 2019-11-11
fix(oas): normalize operation tags

P0lip:
Fixes https://github.com/stoplightio/studio-internal/issues/998


src/oas3/operation.ts
```diff
@@ -43,7 +43,7 @@ export const transformOas3Operation: Oas3HttpOperationTransformer = ({ document,
       operation.requestBody as RequestBodyObject,
     ),
     callbacks: operation.callbacks && translateToCallbacks(operation.callbacks),
-    tags: translateToTags(operation.tags || []),
+    tags: translateToTags(getOasOperationTags(operation)),
```
marbemac: Why expand size of object being passed in, all we need are the tags right? Rather than the entire operation object I mean.

PR #46, opened 2019-11-29
fix(oas2): handle invalid server host

P0lip:
Addresses https://github.com/stoplightio/studio-internal/issues/1363

### DIFF
--- a/src/oas2/transformers/__tests__/servers.test.ts
+++ b/src/oas2/transformers/__tests__/servers.test.ts
@@ -1,19 +1,20 @@
 import { translateToServers } from '../servers';
 
+type GlobalWithLocation = typeof global & { location: Partial<Location> & { href: string } };
+
 describe('translateToServers', () => {
+  afterAll(() => {
+    delete (global as GlobalWithLocation).location;
+  });
+
   // Assures: https://stoplightio.atlassian.net/browse/SL-976
   test('given executed in a browser context and a query in location.href should not inherit that query', () => {
-    // @ts-ignore - ignore the fact that `location` does not exist on global object. Mimic browser.
-    const loc = global.location;
-    // @ts-ignore - ignore the fact that `location` does not exist on global object. Mimic browser.
-    global.location = {
+    (global as GlobalWithLocation).location = {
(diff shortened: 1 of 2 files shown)


src/oas2/transformers/__tests__/servers.test.ts
```diff
@@ -94,4 +99,21 @@ describe('translateToServers', () => {
       },
     ]);
   });
+
+  test('should handle invalid server host gracefully', () => {
+    expect(translateToServers({ host: 123 as any, basePath: '/base-path' }, { schemes: ['http', 'https'] })).toEqual(
```
lottamus: Why not handle it as a string?

```ts
[
  {
    url: 'http://123'
  }
]
```

marbemac: I think OK to just toss it - that would be a very odd server host ha.

src/oas2/transformers/servers.ts
```diff
@@ -20,8 +20,8 @@ export function translateToServers(spec: Partial<Spec>, operation: Partial<Opera
       .scheme(scheme)
       .host(host);
 
-    if (basePath) {
-      uri = uri.path(basePath);
+    if (typeof basePath === 'string' && basePath.length > 0) {
+      uri = uri.path(String(basePath));
```
marbemac: If `basePath` is already a string, why the cast here?

src/oas2/transformers/__tests__/servers.test.ts
lottamus: Users do very _odd_ things with their OpenAPI and not much extra effort to support it lol 🤷‍♂

XVincentX: From a Prism perspective, we do not use servers anyway so whatever works for you will work for us as well.

src/oas2/transformers/servers.ts
P0lip:
```suggestion
      uri = uri.path(basePath);
```
Yeah, it's a leftover, as I hadn't had the check before I added it.

marbemac:
Good to go @P0lip!

PR #49, opened 2019-12-16
Preserve readOnly, writeOnly, xml and externalDocs in schemas

karol-maciaszek:
Add support for unsupported schema properties: `readOnly, writeOnly, xml and externalDocs`

### DIFF
--- a/src/oas2/transformers/params.ts
+++ b/src/oas2/transformers/params.ts
@@ -59,14 +59,11 @@ export function translateToHeaderParam(parameter: HeaderParameter): IHttpHeaderP
 }
 
 export function translateToHeaderParams(headers: { [headerName: string]: Header }): IHttpHeaderParam[] {
-  return map(headers, (header, name) => {
-    const param: IHttpHeaderParam = {
-      ...buildSchemaForParameter(Object.assign({ name }, header)),
-      name,
-      style: HttpParamStyles.Simple,
-    };
-    return param;
-  });
+  return map(headers, (header, name) => ({
+    ...buildSchemaForParameter(Object.assign({ name }, header)),
+    name,
+    style: HttpParamStyles.Simple,
+  }));
 }
 
 export function translateToBodyParameter(body: BodyParameter, consumes: string[]): IHttpOperationRequestBody {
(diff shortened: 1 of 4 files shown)


src/oas2/transformers/params.ts
```diff
@@ -179,7 +179,7 @@ export function translateToPathParameter(parameter: PathParameter): IHttpPathPar
 }
 
 function buildSchemaForParameter(
-  param: QueryParameter | PathParameter | HeaderParameter | FormDataParameter | Header,
+  param: (QueryParameter | PathParameter | HeaderParameter | FormDataParameter | Header) & { readOnly?: boolean },
```
karol-maciaszek: @XVincentX this is due to a typing problem. I haven't analyzed in thoroughly yet but I guess somewhere `required` is defined as array and elsewhere as boolean. When I add `readOnly`, TS compiler picks different type and complains. 

I'll take a closer look tomorrow.

philsturgeon: "required is an array but elsewhere its a boolean" 

This is not the same keyword. Inside a Schema Object, required is an array. Parameter Object (not a schema object) required is a boolean. We cannot mix up Schema Objects and Parameter Objects, or get confused about which keyword goes in where. The spec should show us which goes where, even if it is a bloomin tricky read.

philsturgeon: A more specific note: `readOnly` is only mentioned in one place in the spec, and its in the Schema Object. This probably should not be here.

src/oas3/transformers/__tests__/content.test.ts
```diff
@@ -207,6 +211,53 @@ describe('translateMediaTypeObject', () => {
       );
       expect(translatedObject.schema).toHaveProperty('deprecated', true);
     });
+
+    test('will keep the writeOnly property', () => {
```
XVincentX: You can group all these tests in a `describe` block so you can transform the object *once* and then do the various `it/expect` statements

src/oas2/transformers/params.ts
XVincentX: Ok this looks weird. I'm digging into this as well and see if I can help.

XVincentX: @karol-maciaszek I do not think this change is required at all, effectively this property is only in the `Schema` object. Tests are indeed passing even without this change.

src/oas3/transformers/__tests__/content.test.ts
karol-maciaszek: Done!

src/oas2/transformers/params.ts
karol-maciaszek: Yes, the code actually copies schema stuff from a param to schema. So `readOnly` is not applicable here.

PR #51, opened 2020-01-07
fix: include schema in parameter check

mallachari:
Addresses: https://github.com/stoplightio/studio-internal/issues/1448#event-2895785126

Newly added response header wasn't treated as oas3 header when building virtual node

Changes:
- include `schema` when checking if object is Parameter (Header)


mallachari:
@lottamus looks that your review is necessary, could you take a look?

PR #52, opened 2020-01-10
fix: filter out null request parameters

P0lip:
Fixes https://github.com/stoplightio/studio-internal/issues/1608

### DIFF
--- a/src/oas/accessors.ts
+++ b/src/oas/accessors.ts
@@ -1,6 +1,6 @@
 import { isObject, unionBy } from 'lodash';
 
-export function getOasParameters<ParamType extends { name: string; in: string }>(
+export function getValidOasParameters<ParamType extends { name: string; in: string }>(
   operationParameters: ParamType[] | undefined,
   pathParameters: ParamType[] | undefined,
 ) {
--- a/src/oas/accessors.ts
+++ b/src/oas/accessors.ts
@@ -1,12 +1,12 @@
-import { unionBy } from 'lodash';
+import { isObject, unionBy } from 'lodash';
 
 export function getOasParameters<ParamType extends { name: string; in: string }>(
   operationParameters: ParamType[] | undefined,
   pathParameters: ParamType[] | undefined,
 ) {
   return unionBy(operationParameters, pathParameters, (parameter?: ParamType) =>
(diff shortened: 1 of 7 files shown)


src/oas3/transformers/request.ts
```diff
@@ -51,6 +51,7 @@ export function translateToRequest(
   };
 
   for (const parameter of parameters) {
+    if (!isObject(parameter)) continue; // just in case (handled by `getOasParameters` already)
```
XVincentX: Hmm, if so — can't we remove it?

src/oas/accessors.ts
```diff
@@ -1,12 +1,12 @@
-import { unionBy } from 'lodash';
+import { isObject, unionBy } from 'lodash';
 
 export function getOasParameters<ParamType extends { name: string; in: string }>(
```
XVincentX: This should probably become `getValidOasParameters` – but we can keep it as it is.

P0lip:
> Looks good but can you do the same of the oas2 parameters?

@lottamus they seem to use the same accessor, hence the oas2 case should be covered as well.
I'll add a test proving that.

src/oas3/transformers/request.ts
```diff
@@ -8,7 +8,7 @@ import {
   IHttpQueryParam,
   IMediaTypeContent,
 } from '@stoplight/types';
-import { map, omit, pickBy } from 'lodash';
+import { isObject, map, omit, pickBy } from 'lodash';
```
lottamus:
```suggestion
import { map, omit, pickBy } from 'lodash';
```

src/oas2/__tests__/operation.test.ts
```diff
@@ -90,4 +91,52 @@ describe('transformOas2Operation', () => {
...
+      paths: {
+        '/users/{userId}': {
+          get: {
+            parameters: [
+              {
+                in: 'header',
+                name: 'name',
+                schema: {
+                  type: 'string',
+                },
```
lottamus: OpenAPI v2 parameters don't have a "schema" property if you want to be technically correct 🤓 

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameters-definitions-object

```ts
type: 'string',
```

PR #54, opened 2020-01-14
fix: handle invalid oauth2 security flows

P0lip:
Should fix https://github.com/stoplightio/studio-internal/issues/1632

### DIFF
--- a/src/oas3/transformers/securities.ts
+++ b/src/oas3/transformers/securities.ts
@@ -1,4 +1,4 @@
-import { DeepPartial, IApiKeySecurityScheme, IOauthFlowObjects } from '@stoplight/types';
+import { DeepPartial, IApiKeySecurityScheme, IOauthFlowObjects, Optional } from '@stoplight/types';
 import { compact, pickBy } from 'lodash';
 import { OAuthFlowsObject, OpenAPIObject, SecuritySchemeObject } from 'openapi3-ts';
 import { getSecurities, OperationSecurities, SecurityWithKey } from '../accessors';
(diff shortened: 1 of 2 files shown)


src/oas3/transformers/securities.ts
```diff
@@ -66,9 +66,13 @@ export function transformToSingleSecurity(
   return undefined;
 }
 
-function transformFlows(flows: DeepPartial<OAuthFlowsObject>): IOauthFlowObjects {
+function transformFlows(flows: Optional<DeepPartial<OAuthFlowsObject>>): IOauthFlowObjects {
   const transformedFlows: IOauthFlowObjects = {};
 
+  if (flows === void 0) {
```
lottamus: Should we also protect against `null` and non-object values too while we're at it?

```ts
    const document: Partial<OpenAPIObject> = {
      components: {
        securitySchemes: {
          t1: { type: 'oauth2', flows: "invalid" },
        },
      },
    };

    const document: Partial<OpenAPIObject> = {
      components: {
        securitySchemes: {
          t1: { type: 'oauth2', flows: null },
        },
      },
    };

    const document: Partial<OpenAPIObject> = {
      components: {
        securitySchemes: {
          t1: { type: 'oauth2', flows: 999 },
        },
      },
    };

    const document: Partial<OpenAPIObject> = {
      components: {
        securitySchemes: {
          t1: { type: 'oauth2', flows: [] },
        },
      },
    };
```

P0lip: Yeah, it could be null, true.

PR #57, opened 2020-01-21
fix: override definition scopes with operation scopes

rainum:
According to [OAS 2 spec](https://swagger.io/docs/specification/2-0/authentication):

> Global security can be overridden in individual operations to use a different authentication type, different OAuth 2 scopes, or no authentication at all.

This PR adds ability to override oauth2 definition scopes with an operation scopes and it should fix https://github.com/stoplightio/platform/issues/1599

### DIFF
--- a/src/oas2/__tests__/accessors.test.ts
+++ b/src/oas2/__tests__/accessors.test.ts
@@ -52,8 +52,8 @@ describe('accessors', () => {
               authorizationUrl: 'http://swagger.io/api/oauth/dialog',
               flow: 'implicit',
               scopes: {
-                'read:pets': '',
-                'write:pets': '',
+                'write:pets': 'modify pets in your account',
+                'read:pets': 'read your pets',
               },
               type: 'oauth2',
               key: 'petstore_auth',
@@ -88,8 +88,8 @@ describe('accessors', () => {
               authorizationUrl: 'http://swagger.io/api/oauth/dialog',
               flow: 'implicit',
               scopes: {
-                'read:pets': '',
-                'write:pets': '',
+                'write:pets': 'modify pets in your account',
+                'read:pets': 'read your pets',
               },
               type: 'oauth2',
               key: 'petstore_auth',
@@ -189,7 +189,7 @@ describe('accessors', () => {
           {
             authorizationUrl: 'http://swagger.io/api/oauth/dialog',
             flow: 'implicit',
-            scopes: { 'write:pets': '' },
+            scopes: { 'write:pets': 'modify pets in your account' },
             type: 'oauth2',
             key: 'petstore_auth',
           },
@@ -212,7 +212,10 @@ describe('accessors', () => {
           {
             authorizationUrl: 'http://swagger.io/api/oauth/dialog',
             flow: 'implicit',
-            scopes: { 'write:pets': '', 'read:pets': '' },
+            scopes: {
+              'write:pets': 'modify pets in your account',
+              'read:pets': 'read your pets',
+            },
             type: 'oauth2',
             key: 'petstore_auth',
           },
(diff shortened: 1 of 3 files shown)


src/oas2/__tests__/accessors.test.ts
```diff
@@ -52,8 +52,8 @@ describe('accessors', () => {
               authorizationUrl: 'http://swagger.io/api/oauth/dialog',
               flow: 'implicit',
               scopes: {
-                'read:pets': 'read your pets',
-                'write:pets': 'modify pets in your account',
+                'read:pets': '',
+                'write:pets': '',
```
lottamus: This doesn't look right. Why do we lose the descriptions? We need those in the UI

rainum: Yeah, I know. But since we're overriding this particular security scheme with

```ts
  const securityFixture: Array<Dictionary<string[], string>> = [
    {
      api_key: [],
      petstore_auth: ['write:pets', 'read:pets'],
    },
  ];
```

there is no way to preserve description because `petstore_auth: ['write:pets', 'read:pets']` array is converted to object.

lottamus: Those scopes are being pulled from the security definition which contains the descriptions. You can look up the definition and attach it to the operation's scope

rainum: sure, but what if the scope is totally different? I'm pretty sure that this will be true for the most cases.

rainum: And this won't be "override" anymore. This looks like merge operation to me. But spec says `security can be overridden`. What do you think?

lottamus: Operations can override the root level ["global" security](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityRequirementObject) object. The operation itself does not define scopes. The scopes are defined by the [security definition](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityDefinitionsObject) object. The global and operation securities choose which of those scopes they require.

rainum: Hm, so you want to say that we need to ignore operation's scopes completely which are defined by  `securityFixture`? That makes no sense to me. At the https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#patterned-fields-13 section it says:

> Each name must correspond to a security scheme which is declared in the Security Definitions. If the security scheme is of type "oauth2", **then the value is a list of scope names required for the execution.** For other security scheme types, the array MUST be empty.

lottamus: > so you want to say that we need to ignore operation's scopes completely

Operations don't define scopes. A [Security Definition](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityDefinitionsObject) with type "oauth2" defines scopes.

An OpenAPI document has a global ["security"](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityRequirementObject) property which applies certain Security Definitions to **ALL** operations within the document.

An [Operation](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-5) can override the global security with it's own "security" property BUT it must be a defined Security Definition.

> The name used for each property MUST correspond to a security scheme declared in the Security Definitions.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-5

> A declaration of which security schemes are applied for this operation. The list of values describes alternative security schemes that can be used (that is, there is a logical OR between the security requirements). This definition overrides any declared top-level security. To remove a top-level security declaration, an empty array can be used.


When transforming the global security and operation security, http-spec maps over the security keys and denormalizes the matching security definition as defined here: https://github.com/stoplightio/types/blob/master/src/http-spec.ts#L39

Does that help?

Basically you're filtering the scopes correctly, but you still need to preserve the definition: https://github.com/stoplightio/types/blob/master/src/http-spec.ts#L206

lottamus: No worries, looks good now!

PR #58, opened 2020-01-21
fix: ignore nullish oas3 responses

P0lip:
Fixes https://github.com/stoplightio/studio-internal/issues/1618

### DIFF
--- a/src/oas3/operation.ts
+++ b/src/oas3/operation.ts
@@ -1,4 +1,5 @@
 import { IHttpOperation } from '@stoplight/types';
+import { IHttpOperationResponse } from '@stoplight/types/dist';
 import { get, isNil, omitBy } from 'lodash';
 import { OpenAPIObject, OperationObject, ParameterObject, PathsObject, RequestBodyObject } from 'openapi3-ts';
 
@@ -38,7 +39,8 @@ export const transformOas3Operation: Oas3HttpOperationTransformer = ({ document,
     method,
     path,
     summary: operation.summary,
-    responses: translateToResponses(operation.responses),
+    // fixme: this doesn't make sense, since the array could be empty
+    responses: translateToResponses(operation.responses) as IHttpOperationResponse[] & { 0: IHttpOperationResponse },
     servers: Array.isArray(servers) ? translateToServers(servers.filter(isServerObject)) : [],
     request: translateToRequest(
       getValidOasParameters(operation.parameters as ParameterObject[], pathObj.parameters),
--- a/src/oas3/guards.ts
+++ b/src/oas3/guards.ts
@@ -1,8 +1,9 @@
 import { Dictionary } from '@stoplight/types';
-import { isObject } from 'lodash';
+import { isObject, isObjectLike } from 'lodash';
 import {
   BaseParameterObject,
   HeaderObject,
+  ResponseObject,
   SecuritySchemeObject,
   ServerObject,
   ServerVariableObject,
@@ -50,3 +51,13 @@ export const isTagObject = (maybeTagObject: unknown): maybeTagObject is TagObjec
 
   return false;
 };
+
+export const isResponseObject = (maybeResponseObject: unknown): maybeResponseObject is ResponseObject =>
+  isObject(maybeResponseObject) &&
+  ('description' in maybeResponseObject ||
+    'headers' in maybeResponseObject ||
+    'content' in maybeResponseObject ||
+    'links' in maybeResponseObject);
+
+export const isDictionary = (maybeDictionary: unknown): maybeDictionary is Dictionary<unknown> =>
+  isObjectLike(maybeDictionary);
--- a/src/oas3/guards.ts
+++ b/src/oas3/guards.ts
@@ -1,5 +1,5 @@
 import { Dictionary } from '@stoplight/types';
-import { isObject, isObjectLike } from 'lodash';
+import { isObject } from 'lodash';
 import {
   BaseParameterObject,
   HeaderObject,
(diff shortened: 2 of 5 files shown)


src/oas3/operation.ts
```diff
@@ -38,7 +39,8 @@ export const transformOas3Operation: Oas3HttpOperationTransformer = ({ document,
     method,
     path,
     summary: operation.summary,
-    responses: translateToResponses(operation.responses),
+    // fixme: this doesn't make sense, since the array could be empty
```
P0lip: @XVincentX we chatted about this on Slack.
I'm leaning towards making a change in `@stoplight/types` to accept any array of `IHttpOperationResponse`, but curious to hear your thoughts.

lottamus: +1 for updating @stoplight/types to be more lenient

P0lip:
> You already know my opinion, I really think this is so wrong, but I do not have the forces to keep fighting this battle.

There is a fundamental issue here.
Prism has by far different needs than Studio (Graphite) does.
Prism expects to operate on a *valid* document, therefore it'd actually benefit from http-spec throwing exceptions. Studio, on the other hand, needs to work on pretty much every single document, therefore the transformation should never fail. 
Http-spec won't satisfy the needs of all consumers unless we provide some sort of strict/loose mode, but this would introduce lots of burden.

XVincentX:
I would say we'll probably depart from Http-Spec pretty soon. I do not want to add a gazillion of checks in Prism.

marbemac:
> I would say we'll probably depart from Http-Spec pretty soon. I do not want to add a gazillion of checks in Prism.

Highly discourage this - we should have a broader conversation to come up with a better solution if changes like the one in this PR are impacting Prism in a negative way. 

However, I'm not sure I understand - how does this change impact Prism negatively? The output of the `http-spec` operation transformers should always be a valid `IHttpOperation` object - Prism should be able to rely on this, and not need to add extra checks anywhere. For example with the change in this PR you no longer need to worry about null (invalid) values being in the responses array.

What we're trying to do with these PRs in http-spec is to make sure we always produce a valid `IHttpOperation` object that prism and other downstream consumers can rely on, while not requiring the end user's specification file to be perfect (because let's face it, they rarely are).

XVincentX:
> The output of the http-spec operation transformers should always be a valid IHttpOperation object

This really comes down what's your idea of a valid `IHttpOperation` and what are the actions to make it valid in case it is not. I'll do two specific examples but you can probably expand it in other scenarios as well:

1. Is an IHttpOperation without at least a response valid? To me it's not. If you can't construct an operation with a response, such response has no reason to exist, and it should not be produced at all. The proposed relaxation says that instead it's just fine. I understand your use case (editing experience) but I really feel we're leaning too much on that direction.

2. What are the actions to make an invalid HttpOperation valid? I personally believe in the "an invalid http operation is not an operation", but the approach that's used here is by dropping invalid parts. Sure, that probably works but then what happens is:

* User opens an issue with an OpenAPI document and Prism responds incorrectly
* I try the document and Prism effectively does not respond
* I run the document on Spectral which says "yeah, the document is ok"
* I do not understand what's going on and I feel a terrible programmer
* I walkup half of the Prism source code to then end up manually inspecting the produced operation array
* I found out invalid content has been dropped
* I need to reply saying "yo your document is not valid"
* Close the issue

---

It would be way better instead if:

* User loads document in Prism
* "YO DOCUMENT INVALID FIX IT"
* End, no issue opened

My point is we're trying to handle two very different use cases with a single piece of software. Clearly, one of these will suffer.

1. Studio: give me everything, I'll figure something out.
2. Prism: give me what I'll ask you or I won't even run.

marbemac:
> This really comes down what's your idea of a valid IHttpOperation

Good question - what I mean by this is that these transformers should always produce a result that implements the `IHttpOperation` interface - and not throw an error or fail during transformation wherever possible. So if prism is built to work with that interface, then the result here should be good to go. 

> we should shame the user as well

Linters should shame the user (like Spectral). Maybe we should put a shame easter egg in Studio :).

> I run the document on Spectral which says "yeah, the document is ok" (which is not true)

Ok well this sounds like a bug. If the document is not valid, we should report it? Does this happen often, and can you open issues when it does?

---

IMHO if we really want prism to require valid specification, we should just run a linter like spectral on startup, and error out if there are issues? This WILL be annoying for many users, since frankly so many people have small issues in their specification, but maybe it's worth it what do you think? Could even add a flag to ignore this check, which would solve most of your workflow since it would be clear that user is ignoring the check (and it's clear to them that they are ignoring and probably have an issue with their spec).

Basically, not sure it's an issue in `http-spec` or requires prism to use another transformer - we might be able to solve the problem with a lint check in prism?

XVincentX:
> So if prism is built to work with that interface, then the result here should be good to go.

I agree, and that is exactly the reason why I have been putting a lot of constraints on Http Spec type to make sure that — if somebody can produce a IHttpOperation object, it is good to go. All these checks have been removed and with the change that has been proposed we're getting to a point where:

```ts
const op: IHttpOperation = {
  id: 'id',
  method: 'ahahca',
  path: 'ajdoasjdi',
  responses: []
}
```

would be a valid HttpOperation. This operation has an invalid method, an invalid path and has no response. There is no way Prism can get a response out of it. Still, it's a valid IHttpOperation. Analyser would include this in its process and even report a _breaking change_ in case it gets removed; a breaking change for a thing that shouldn't have been there in the first place.

I also see viable for some other reasons that in the future fields will start to be optional as well and even an empty object would be a valid `IHttpOperation`.

```ts
const op: IHttpOperation = {}
```

What are we supposed to do with such object?

---

Again, I totally understand the Studio use case. 

These invalid objects are probably coming from a user on Studio clicking around and that's still working on the document, and you still want to save its input. It's legit. 

Studio and Spectral work on the source code, Prism and Analyser should work on artefacts. 

Maybe Studio should be working on a `DeepPartial<IHttpOperation>` (or whatever is better for you), and then try to parse the data to `IHttpOperation`.

> frankly so many people have small issues in their specification

Maybe it's time to tell them they've been wrong for ages.

> IMHO if we really want prism to require valid specification, we should just run a linter like spectral on startup, and error out if there are issues?

Why do we need to run an external validator when we can effectively put interface constraints and parse the payload in the correct way in the first instance?

PR #60, opened 2020-01-21
fix: ignore nullish media objects

P0lip:
Fixes https://github.com/stoplightio/studio-internal/issues/1612

### DIFF
--- a/src/oas3/guards.ts
+++ b/src/oas3/guards.ts
@@ -3,6 +3,7 @@ import { isObject } from 'lodash';
 import {
   BaseParameterObject,
   HeaderObject,
+  MediaTypeObject,
   SecuritySchemeObject,
   ServerObject,
   ServerVariableObject,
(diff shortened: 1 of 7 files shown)


src/oas3/guards.ts
```diff
@@ -50,3 +51,10 @@ export const isTagObject = (maybeTagObject: unknown): maybeTagObject is TagObjec
 
   return false;
 };
+
+export const isMediaObject = (maybeMediaObject: unknown): maybeMediaObject is MediaTypeObject =>
+  isObject(maybeMediaObject) &&
+  ('schema' in maybeMediaObject ||
+    'examples' in maybeMediaObject ||
+    'example' in maybeMediaObject ||
+    'encoding' in maybeMediaObject);
```
lottamus: According to @stoplight/types "schema", "examples", and "encodings" are all optional on the media type content interface. So we probably shouldn't require them here, right?

https://github.com/stoplightio/types/blob/master/src/http-spec.ts#L128-L136

XVincentX: @lottamus is right, according to the spec the Media Object has no required field — making fundamentally *anything* a valid `MediaObject`

P0lip: Yeah, that's right, all properties are optional.
That being said, if none of them is present, we won't produce anything meaningful anyway.

PR #61, opened 2020-01-23
Postman Collection - Request

karol-maciaszek:
Introduces complete implementation of Request transformations for Postman Collections. Test coverage of new features is 100%. I also added missing tests to `traverseItemsAndGroups`.

**Note2**: temporarily setting base to previous PR to show the differences


src/postman/operation.ts
```diff
@@ -1,3 +1,81 @@
...
+      header,
+      path: transformPathParams(item.request.url.path),
+      body: item.request.body ? transformBody(item.request.body, findContentType(item.request.headers)) : undefined,
+    },
+    responses: [],
+    /*
+    servers: ...,
+    security: ...,
+    */
+  } as any;
```
karol-maciaszek: I know

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+  IHttpQueryParam,
+  IMediaTypeContent,
+} from '@stoplight/types';
+import { JSONSchema4 } from 'json-schema';
+import { FormParam, HeaderDefinition, PropertyList, QueryParam, RequestBody } from 'postman-collection';
+import { transformDescriptionDefinition, transformPostmanTemplate, transformValueToHttpParam } from '../util';
+
+export function transformQueryParam(queryParam: QueryParam): IHttpQueryParam {
+  return {
+    name: queryParam.key as string, // no key no game
```
XVincentX: This can effectively be null according to the spec, I am afraid we can't really get it in this way.

The Postman collection apparently translates/sends to the network a null parameter as an empty string — we should be able to do that here as well, although we might need to group them

```suggestion
    name: queryParam.key || ''
```

src/postman/util.ts
```diff
@@ -1,4 +1,50 @@
...
+      return { type: 'integer' };
+  }
+
+  return;
+}
+
+export function transformValueToHttpParam(value: string): Pick<IHttpParam, 'schema' | 'examples'> {
+  const schema = transformValueTemplateToSchema(value);
+  if (schema) return { schema };
+
```
XVincentX: I guess this part needs to be revisited, right?

src/postman/util.ts
```diff
@@ -1,4 +1,50 @@
...
+    case '<string>':
+      return { type: 'string' };
+    case '<long>':
+      return { type: 'integer' };
+  }
+
+  return;
+}
+
+export function transformValueToHttpParam(value: string): Pick<IHttpParam, 'schema' | 'examples'> {
```
XVincentX: This function is not really returning an Http Param, it's rather extracting an example or returning a schema — can we reword the name?

package.json
```diff
@@ -33,11 +33,13 @@
     "@stoplight/json": "^3.4.0",
     "@stoplight/types": "^11.4.0",
     "@types/swagger-schema-official": "~2.0.20",
+    "@types/to-json-schema": "^0.2.0",
     "@types/urijs": "~1.19.4",
     "lodash": "^4.17.15",
     "@openapi-contrib/openapi-schema-to-json-schema": "^3.0",
     "openapi3-ts": "~1.3.0",
     "postman-collection": "^3.5.5",
+    "to-json-schema": "^0.2.3",
```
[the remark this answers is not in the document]
karol-maciaszek: I couldn't get quicktype to produce additionalOptions: true. It is also quite big (1.6mb). My suggestion is to stay with `to-json-schema` at least for now. If it happens that somehow it is not a fit, it is just a simple function call with single argument being an object. We can replace it fairly easy.

src/postman/operation.ts
```diff
@@ -1,3 +1,81 @@
...
+    description: item.description && transformDescriptionDefinition(item.description),
+    method,
+    path,
+    summary: item.name,
+    request: {
+      query: item.request.url.query.all().map(transformQueryParam),
+      header,
+      path: transformPathParams(item.request.url.path),
+      body: item.request.body ? transformBody(item.request.body, findContentType(item.request.headers)) : undefined,
+    },
```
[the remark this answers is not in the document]
karol-maciaszek: Done, added `transformRequest` func.

src/postman/operation.ts
karol-maciaszek: Yes, `responses[]` requires at least one response so I needed to cast it and yolo1!!

src/postman/transformers/request.ts
karol-maciaszek: I checked in Postman App: when I make key of query param an empty string, it is not added to the request at all. It is also exported as empty string.

Thinking about two things:
1. leave a param empty-string named
2. remove those empty-string named params

Quite hard for me to make a decision:
- it can be useful for someone to keep the information about those params (e.g. one can fill missing information in our eco-system)
- keeping empty-named params will make then unusable in that form? it may cause problems with creating requests or mocking?

@philsturgeon @XVincentX WDYT?

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+    style: HttpParamStyles.Form,
+    ...(queryParam.value ? transformValueToHttpParam(queryParam.value) : undefined),
+  };
+}
+
+export function transformHeader(header: HeaderDefinition): IHttpHeaderParam {
+  return {
+    name: header.key as string,
```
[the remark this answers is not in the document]
karol-maciaszek: Hm, good point. I misused the type here. `HeaderDefinition` reflects the types of JSON documents. `Header` is runtime representation and should be used here. Thanks!

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+      };
+
+    case 'urlencoded':
+      if (!body.urlencoded) return;
+      return {
+        contents: [transformParamsBody<QueryParam>(body.urlencoded, mediaType || 'application/x-www-form-urlencoded')],
+      };
+  }
+
+  return;
```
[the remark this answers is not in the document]
karol-maciaszek: 1. since http-spec is used as a library we should probably establish a channel for communicating warnings. Simple `console.warn` could make too much noise (thinking about platform logs). On the other hand, if there is noise, we should implement missing mode.
2. the action if there is no corresponding-to-mode key is skipping entire body. It is hard to say whether such case is a result of error or expected case, but our reaction seems to be OK

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function transformRawBody(raw: string, mediaType?: string): IMediaTypeContent {
+  try {
+    const parsed = JSON.parse(raw);
```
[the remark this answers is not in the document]
karol-maciaszek: Refactored so now it tries to parse only if mediaType is JSON-ish. Added `type-is` dependency for that.

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+        {
+          key: 'default',
+          value: parsed,
+        },
+      ],
+      schema: transformPostmanTemplate(parsed),
+    };
+  } catch (e) {
+    /* noop */
+  }
```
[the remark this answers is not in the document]
karol-maciaszek: Do you mean a case when `mediaType=application/json` and JSON is non-parseble?
Currently, it returns plain-text example and `application/json` mediaType.

Hard to say if we should override user's decision (aware or not) of having JSON mediaType and non-parseable body. We also can't protect prism against such input - user is still able to provide such non-sense in the OAS spec.

I'm leaning towards leaving it like that.

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  return {
+    mediaType: mediaType || 'text/plain',
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+      if (!body.urlencoded) return;
+      return {
+        contents: [transformParamsBody<QueryParam>(body.urlencoded, mediaType || 'application/x-www-form-urlencoded')],
+      };
+  }
+
+  return;
+}
+
+function transformRawBody(raw: string, mediaType?: string): IMediaTypeContent {
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

src/postman/util.ts
karol-maciaszek: Hm, you right it is not HttpParam. Hard to say what it is but `..Schema` is a better name. Updated.

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+  params: PropertyList<T>,
+  mediaType: string,
+): IMediaTypeContent {
+  return {
+    mediaType,
+    schema: {
+      type: 'object',
+      properties: params.all().reduce<NonNullable<JSONSchema4['properties']>>((props, item) => {
+        // @todo: !
+        props[item.key!] = {
```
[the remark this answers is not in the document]
karol-maciaszek: I don't think it makes logical sense. Basically we are constructing an example, so for sure the id is not one of the keys. What about `|| ''`?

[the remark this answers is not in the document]
karol-maciaszek: We should be compatible with https://github.com/stoplightio/http-spec/pull/61/files#r372344192

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,121 @@
...
+          type: 'string',
+          description: item.description && transformDescriptionDefinition(item.description),
+        };
+        return props;
+      }, {}),
+    },
+    examples: [
+      {
+        key: 'default',
+        value: params.all().reduce((value, item) => {
```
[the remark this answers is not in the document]
karol-maciaszek: Fixed 👍

src/postman/util.ts
```diff
@@ -1,4 +1,50 @@
-import { Item, ItemGroup } from 'postman-collection';
+import { IHttpParam } from '@stoplight/types';
+import { JSONSchema4 } from 'json-schema';
+import { DescriptionDefinition, Item, ItemGroup } from 'postman-collection';
+import * as toJsonSchema from 'to-json-schema';
+import { JSONSchema3or4 } from 'to-json-schema';
```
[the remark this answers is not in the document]
philsturgeon: Yeah this has to be 4 or 5 (they're kinda the same thing so we can get away with it) but lets not mention 3 anywhere.

src/postman/__tests__/operation.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        expect(
+          transformPostmanCollectionOperation({
+            document: {
+              item: [
+                {
+                  request: {
+                    method: 'get',
+                    url: '/path/:param?a=b',
+                    body: { mode: 'raw', raw: 'test' } as RequestBody,
+                    header: [{ key: 'header', value: 'a header' }] as HeaderDefinition,
```
XVincentX: Hmm, I'll try to check out if we can get rid of this. @karol-maciaszek how did you generate this sample document? Did you use Postman?

src/postman/__tests__/operation.test.ts
```diff
@@ -0,0 +1,89 @@
...
+                  description: 'desc',
+                },
+              ],
+            },
+            method: 'get',
+            path: '/path/{param}',
+          }),
+        ).toEqual({
+          description: 'desc',
+          id: '?http-operation-id?',
```
XVincentX: @marbemac Can we get rid of this property? It has been there since forever and it's completely useless at the moment.

src/postman/__tests__/operation.test.ts
```diff
@@ -0,0 +1,89 @@
...
+          iid: expect.stringMatching(/^[a-z0-9]{8}-([a-z0-9]{4}-){3}[a-z0-9]{12}$/),
+          method: 'get',
+          path: '/path/{param}',
+          request: expect.any(Object),
```
XVincentX: Hmm, can't we be a little bit more precise about this assertion?

src/postman/__tests__/operation.test.ts
```diff
@@ -0,0 +1,89 @@
...
+          responses: [],
```
XVincentX: Is this temporary?

src/postman/__tests__/operation.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    describe('description is not set', () => {
+      it('returns operation without description', () => {
+        expect(
+          transformPostmanCollectionOperation({
+            document: { item: [{ request: { method: 'get', url: '/path' } }] },
+            method: 'get',
+            path: '/path',
+          }),
+        ).toEqual(expect.objectContaining({ description: undefined }));
```
XVincentX: `expect.not.toHaveProperty`?
```suggestion
        ).not.toHaveProperty('description');
```

src/postman/__tests__/operation.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      it('returns operation with body media-type set', () => {
+        expect(
+          transformPostmanCollectionOperation({
+            document: {
+              item: [
+                {
+                  request: {
+                    method: 'get',
+                    url: '/path',
+                    header: [{ key: 'content-type', value: 'application/json' }] as HeaderDefinition,
```
XVincentX: ditto, I'll check out if I can get rid of this cast

src/postman/transformers/__tests__/request.test.ts
```diff
@@ -0,0 +1,351 @@
...
+    ]);
+  });
+});
+
+describe('transformBody()', () => {
+  describe('body is passed in raw mode', () => {
+    describe('body is defined', () => {
+      describe('mediaType is JSON-ish', () => {
+        describe('body is correctly defined json', () => {
+          expect(transformBody(new RequestBody({ mode: 'raw', raw: '{"a":"b"}' }), 'application/nice+json')).toEqual({
```
XVincentX: There's a missing `it` statement with the intention, such as `it('creates a body definition object', ()=>)

src/postman/transformers/__tests__/request.test.ts
```diff
@@ -0,0 +1,351 @@
...
+                  properties: { a: { type: 'string' } },
+                  type: 'object',
+                },
+              },
+            ],
+          });
+        });
+
+        describe('body is not a correct JSON', () => {
+          expect(transformBody(new RequestBody({ mode: 'raw', raw: '"a":"b"' }), 'application/json')).toEqual({
```
XVincentX: ditto, missing `it` statement

src/postman/transformers/__tests__/request.test.ts
```diff
@@ -0,0 +1,351 @@
...
+        });
+
+        describe('media type is not defined', () => {
+          it('returns body containing example and text/plain media type', () => {
+            expect(transformBody(new RequestBody({ mode: 'raw', raw: "I'm a goat. Bleeet!" }))).toEqual({
+              contents: [
+                {
+                  examples: [{ key: 'default', value: "I'm a goat. Bleeet!" }],
+                  mediaType: 'text/plain',
+                },
```
XVincentX: This case should not technically be possible but I am ok with the fallback.

src/postman/transformers/__tests__/request.test.ts
```diff
@@ -0,0 +1,351 @@
...
+                  properties: {
+                    k1: { type: 'string', description: undefined },
+                    k2: { type: 'string', description: 'd2' },
+                  },
+                },
+              },
+            ],
+          });
+        });
+      });
```
XVincentX: This whole part looks pretty much identical; although it's not critical it would be beneficial to use `describe.each` — this'd have saved some time during the code review.

src/postman/__tests__/operation.test.ts
karol-maciaszek: Initially, I used a tool from here https://github.com/postmanlabs/openapi-to-postman

But Postman app also generates it like that:
- `header` is supposed to be single HeaderDefinition while in reality, it is an array of such
- `body` is supposed to be RequestBody while in reality, it is a RequestBodyDefinition

Here is a screenshot of what has been generated by Postman app:
Screen Shot 2020-02-05 at 12 50 14

I strongly believe those are just bugs in typings (e.g. the contributor saw "header" so assumed it has to be single one). I think when we finish the implementation we need to search for casts along our lines and create a fat PR to postman repo.

src/postman/__tests__/operation.test.ts
karol-maciaszek: +1

src/postman/__tests__/operation.test.ts
karol-maciaszek: Well, I decided not to verify the whole request object because it is huge and it is already verified precisely together with `transformRequest` test. However, I will not defend that. WDYT? Test it all?

src/postman/__tests__/operation.test.ts
karol-maciaszek: Yes, it is. Now I'm working on the code which will fill this up.

src/postman/__tests__/operation.test.ts
karol-maciaszek: 👍

src/postman/__tests__/operation.test.ts
karol-maciaszek: Aye, see here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/postman-collection/index.d.ts#L434 and compare it with what is exported via Postman app.

src/postman/operation.ts
```diff
@@ -1,3 +1,69 @@
...
+
+function findItem({
+  method,
+  path,
+  collection,
+}: {
+  method: string;
+  path: string;
+  collection: Collection;
+}): Item | undefined {
```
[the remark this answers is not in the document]
karol-maciaszek: Feels lighter, thanks!

src/postman/operation.ts
```diff
@@ -1,3 +1,69 @@
...
+      found = item;
+    }
+  });
+
+  return found;
+}
+
+function getPath(url: Url) {
+  return '/' + url.path.map(segment => (segment.startsWith(':') ? `{${segment.substring(1)}}` : segment)).join('/');
+}
```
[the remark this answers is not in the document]
karol-maciaszek: This function rewrites `url.path = `['user', ':name']` into `/user/{name}` which is the format of `path` arg in `transformPostmanCollectionOperation`. Does it explains?

src/postman/transformers/__tests__/request.test.ts
karol-maciaszek: Good catch, thanks

src/postman/transformers/__tests__/request.test.ts
karol-maciaszek: 👍

src/postman/transformers/__tests__/request.test.ts
karol-maciaszek: Well, it is:
Screen Shot 2020-02-05 at 14 26 33

In the Postman app, first I created a normal request and the explicitly removed `Content-type` header.

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,138 @@
...
+  return {
+    query: request.url.query.all().map(transformQueryParam),
+    headers: request.headers.all().map(transformHeader),
+    path: transformPathParams(request.url.path),
+    body: request.body ? transformBody(request.body, findContentType(request.headers)) : undefined,
+  };
+}
+
+function findContentType(headers: HeaderList) {
+  const header = headers.all().find(h => h.key.toLowerCase() === 'content-type');
```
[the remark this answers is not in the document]
karol-maciaszek: Good point, I switched to `headers.get` + `?.toLowerCase()`

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,138 @@
...
+  return header ? header.value.toLowerCase() : undefined;
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,138 @@
...
+function transformParamsBody<T extends FormParam | QueryParam>(
+  params: PropertyList<T>,
+  mediaType: string,
+): IMediaTypeContent {
+  return {
+    mediaType,
+    schema: {
+      type: 'object',
+      properties: params.all().reduce<NonNullable<JSONSchema4['properties']>>((props, item) => {
+        props[item.key || ''] = {
```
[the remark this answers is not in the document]
karol-maciaszek: I'm really not sure whether we should generate artificial ids. This case actually have no good solutions:
1) artificial: user won't understand that, will omit that, will be surprised (chance: always when empty key is used)
2) allow keys to be overwritten: user will be surprised that some params has not been migrated (chance: >1 keys are empty)
3) ignore those with empty keys: user loses data (chances: alway when an empty key is used)

In my opinion, it would be more perfect to do 3). However, it sounds more user-friendly to do 1) with sth link `_generated_beef11`.

Ok, I convinced myself that 1) is good ;)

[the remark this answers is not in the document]
karol-maciaszek: Done! Tell me WDYT about it?

src/postman/types.ts
```diff
@@ -1,7 +1,8 @@
...
+import { HttpOperationTransformer } from '../types';
 
-export type PostmanCollectionTransformOperationOpts = ITransformOperationOpts<Collection>;
-export type PostmanCollectionHttpOperationTransformer = HttpOperationTransformer<
-  PostmanCollectionTransformOperationOpts
->;
+export type PostmanCollectionHttpOperationTransformer = HttpOperationTransformer<{
+  document: CollectionDefinition;
+  path: string;
+  method: string;
```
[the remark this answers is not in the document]
karol-maciaszek: I don't fully understand what you mean?

src/postman/transformers/__tests__/request.test.ts
karol-maciaszek: I'm keeping in mind that `describe.each` exists, however, this time it brings no real value IMO. The only one -we-are-sure-it-is-shared part is `expect...toEqual`. The guts may be always anything. Secondly, inside each `describe` there is most of the time only 2-3 inner tests/describes.

I personally don't feel it will help us here.

src/postman/__tests__/util.test.ts
```diff
@@ -0,0 +1,49 @@
+import { ItemGroup } from 'postman-collection';
+import { transformDescriptionDefinition, transformValueToSchema, traverseItemsAndGroups } from '../util';
+
+describe('transformValueToSchema()', () => {
+  it('returns param with schema and example', () => {
+    expect(transformValueToSchema('test')).toEqual({
```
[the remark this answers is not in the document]
karol-maciaszek: I don't think we should do that. The logic of turning different types of values into schemas is responsibility of `to-json-schema` library. We should not test it (we assume that it is tested and it works 😁). This test is meant only to verify our wrapping code. Which has no branches so a single test is enough.

But! You made me notice that `to-json-schema` is useless here. The func accepts string as arg anyway, so the schema is fixed.

src/postman/__tests__/operation.test.ts
XVincentX: Agreed, let's take note of that and we'll do such thing.

src/postman/__tests__/operation.test.ts
XVincentX: @marbemac ping

src/postman/__tests__/operation.test.ts
XVincentX: Hmm, hmm…I understand the reasoning, I guess we can get away with it.

src/postman/__tests__/util.test.ts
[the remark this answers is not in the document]
XVincentX: We should rename it to `transformStringValueToSchema` then

src/postman/types.ts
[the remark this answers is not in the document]
XVincentX: I made a mistake here, you can ignore this.

src/postman/util.ts
```diff
@@ -1,4 +1,21 @@
-import { Item, ItemGroup } from 'postman-collection';
+import { IHttpParam } from '@stoplight/types';
+import { DescriptionDefinition, Item, ItemGroup } from 'postman-collection';
+
+export function transformValueToSchema(value: string): Pick<IHttpParam, 'schema' | 'examples'> {
```
XVincentX:
```suggestion
export function transformStringValueToSchema(value: string): Pick<IHttpParam, 'schema' | 'examples'> {
```

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,150 @@
...
+          return values;
+        }, {}),
+      },
+    ],
+  };
+}
+
+export function transformRequest(request: Request): IHttpOperationRequest {
+  return {
+    query: request.url.query.all().map(transformQueryParam),
```
XVincentX:
```suggestion
    query: request.url.query.map(transformQueryParam),
```

src/postman/transformers/request.ts
```diff
@@ -0,0 +1,150 @@
...
+    headers: request.headers.all().map(transformHeader),
```
XVincentX:
```suggestion
    headers: request.headers.map(transformHeader),
```

src/postman/transformers/request.ts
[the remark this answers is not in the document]
XVincentX: Yeah that's an hard call, in any case we're returning something anyway, we can start with this.

src/postman/transformers/request.ts
karol-maciaszek: 👍

src/postman/transformers/request.ts
karol-maciaszek: 👍

src/postman/util.ts
karol-maciaszek: 👍

PR #65, opened 2020-02-10
Postman Collection - Security

karol-maciaszek:
Adds support of security in operation.

Depending on the compatibility between HttpOperation/OpenAPI and Postman Collection's RequestAuth, the result can be stored inside security, request.headers or request.query.


src/postman/operation.ts
```diff
@@ -26,18 +32,37 @@ export const transformPostmanCollectionOperation: PostmanCollectionHttpOperation
...
     method,
     path,
     summary: item.name,
-    request: transformRequest(item.request),
+    request,
     responses: item.responses.map(transformResponse),
+    security,
     /*
     servers: ...,
-    security: ...,
```
XVincentX: Good to see with every PR one of these weird pieces going away :)

src/postman/operation.ts
```diff
@@ -57,6 +82,15 @@ function findItem(collection: Collection, method: string, path: string): Item |
   return found;
 }
 
+// this func ensures that both this securityScheme and the one in HttpService share exactly the sme keys
+function findPostmanSecurityScheme(auth: RequestAuth, collection: Collection) {
+  const securityScheme = transformSecurityScheme(auth, () => '1');
+  if (!securityScheme) return;
+
+  const allSecuritySchemes = transformSecuritySchemes(collection);
```
XVincentX: Hmm, computing this every time we need to find a security scheme seems at glance highly inefficient.

src/postman/operation.ts
```diff
@@ -57,6 +82,15 @@ function findItem(collection: Collection, method: string, path: string): Item |
   return found;
 }
 
+// this func ensures that both this securityScheme and the one in HttpService share exactly the sme keys
+function findPostmanSecurityScheme(auth: RequestAuth, collection: Collection) {
```
XVincentX: I do not fully understand why it is required to transform the security Schema to find it in the document, can you clarify?

src/postman/operation.ts
```diff
@@ -26,18 +32,37 @@ export const transformPostmanCollectionOperation: PostmanCollectionHttpOperation
...
+  const request = transformRequest(item.request);
+  const security: HttpSecurityScheme[][] = [];
+
+  if (postmanSecurity) {
+    switch (postmanSecurity.type) {
+      case 'securityScheme':
+        security.push([postmanSecurity.securityScheme]);
+        break;
+      case 'headerParams':
+        request.headers!.push(...postmanSecurity.headerParams);
```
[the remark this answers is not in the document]
karol-maciaszek: You got me. It will be always set because of how `transformRequest` above works. Meaning, it always returns an array.

But you right, I should act like I don posses that knowledge and verify the existence of `headers` and `query`.

src/postman/operation.ts
karol-maciaszek: Yeah :)

src/postman/operation.ts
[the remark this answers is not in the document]
XVincentX: > You got me. It will be always set because of how transformRequest above works. Meaning, it always returns an array.

Hmm, if `transformRequest` gives you these guarantees, we can just modify it to return the correct type, with the property marked as `NonNullable`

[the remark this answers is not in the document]
XVincentX: https://github.com/piotrwitek/utility-types#requiredt-k

src/postman/operation.ts
karol-maciaszek: The `RequestAuth` objects in Postman Collection are always stored inside `item`s. There is no referencing mechanism similar to that found in OpenAPI documents.

That means that there are possibly lots of duplicates. 

1) The `transformSecuritySchemes` basically converts all RequestAuths into SecuritySchemes (almost true), deduplicates them and assigns identifiers. Identifiers contain a sequence number, e.g. `http-0`, `http-1`, `hawk-0`.

2) we have a service transformer, which contains all security schemes

3) the above mechanism ensures that whenever you convert just a single operation or the whole service, the identifiers will be stable.

src/postman/operation.ts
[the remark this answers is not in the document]
karol-maciaszek: I tried using the built-in utility types and I failed. These fill some gaps, thanks!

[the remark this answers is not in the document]
XVincentX: That's fine, these are just types so they do not even get into the final bundle — augmenting the size or something

src/postman/operation.ts
XVincentX: Ok understood.

Looking at the implementation though., I think we should tackle this from a different angle.

Is there any way we can probably pre-compute all the security ahead of the time so we can pass them as a parameter here and use it to look for the security definition we are looking for?

karol-maciaszek: Yeah, I thought about different options but none of them had a significant advantage over this one.

- we **have** to loop over every security in the doc to obtain the id
- we **have** to take care of auth data to perform the deduplication

Having that stated, the available options are:
1. generate IDs independently of transformation
+ faster than generating id & transforming
- not _that_ faster, since we skip only one, quite simple step
- we add complexity to the code (separate id generation)
- the functionality would be possibly duplicated, as it handy to generate IDs while transforming

2. generate IDs during transformation
- slower than 1. since it needs to also transform all entities
+ not _that_ slow, transformation is really simple
+ simple code, no duplication, good maintainability

3. cache!
This solution is an enhancement of 2. The idea would be to locally cache the auth IDs somehow. We know that for the same document the auths will be the same. The calling code could have control over how and when the cache is cleared. I'm thinking about passing somehow a context during the processing of the document (to keep the code functional-ready).

Moreover, 3. can be implemented when we run into actual optimization problems.

XVincentX: Ok let's move forward for now, but I think we're going to need to revisit this at some point.

PR #67, opened 2020-02-11
Postman Collection - Server

karol-maciaszek:
Adds support for servers in operation.


src/postman/transformers/server.ts
```diff
@@ -0,0 +1,10 @@
+import { IServer } from '@stoplight/types/dist';
```
XVincentX:
```suggestion
import { IServer } from '@stoplight/types';
```

src/postman/transformers/server.ts
```diff
@@ -0,0 +1,10 @@
...
+import { Url } from 'postman-collection';
+
+export function transformServer(url: Url): IServer | undefined {
+  if (!url.host) return;
+
+  return {
+    url: `${url.protocol}://${url.host.join('.')}${url.port ? `:${url.port}` : ''}`,
```
XVincentX: Can we use `url.format`? It's going to be way safer

src/postman/transformers/server.ts
```diff
@@ -0,0 +1,10 @@
+import { IServer } from '@stoplight/types/dist';
+import { Url } from 'postman-collection';
+
+export function transformServer(url: Url): IServer | undefined {
+  if (!url.host) return;
```
XVincentX: I'm really wondering why didn't they make this required.

XVincentX: According to the spec the host can be a string or an array of strings — is the SDK taking care of this automatically?

src/postman/transformers/server.ts
karol-maciaszek: 👍

src/postman/transformers/server.ts
karol-maciaszek: The SDK is taking care of conversion `string|string[] => string[]` in that case.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/postman-collection/index.d.ts#L616 - that's how `Url` is defined - it's an `string[]`
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/postman-collection/index.d.ts#L605 - `UrlDefinition` - here we deal with `string | string[]`

Ofc since those types are known to have bugs I double-checked it with tests and this time it acts like it was defined.


---

But you spotted another possible bug: protocol might be undefined. I fixed that already.

src/postman/transformers/server.ts
karol-maciaszek: Done, now `urijs` adds some protection.

XVincentX: Well yeah, at least it moves the responsibility to somebody else :)

PR #68, opened 2020-02-12
Postman Collection - Operations

karol-maciaszek:
Transformation of all operations in given postman collection document.

**Note:** the Postman Collection feature will be divided into several reviewable parts


src/postman/__tests__/operation.test.ts
```diff
@@ -283,3 +283,40 @@ describe('transformPostmanCollectionOperation()', () => {
...
+        expect.objectContaining({
+          path: '/a/a/b',
+        }),
+        expect.objectContaining({
+          path: '/a',
+        }),
+      ]),
+    );
+
+    expect(operations).toHaveLength(4);
```
XVincentX: This statement is already implicit in the first expect. Fine to leave it, just saying.

src/postman/__tests__/util.test.ts
```diff
@@ -32,7 +32,7 @@ describe('traverseItemsAndGroups()', () => {
 
       traverseItemsAndGroups(new ItemGroup({ item: [{ item: [{}] }] }), mockItemCallback, mockGroupItemCallback);
 
-      expect(mockItemCallback).toHaveBeenCalledTimes(2);
+      expect(mockItemCallback).toHaveBeenCalledTimes(1);
```
XVincentX: Why is this being modified?

src/postman/util.ts
```diff
@@ -24,5 +24,4 @@ export function traverseItemsAndGroups(
 ) {
   itemGroup.forEachItem(itemCallback);
   if (itemGroupCallback) itemGroup.forEachItemGroup(itemGroupCallback);
-  itemGroup.forEachItemGroup(ig => traverseItemsAndGroups(ig, itemCallback, itemGroupCallback));
```
XVincentX: Why is the traversing code changing? Does not really add up

src/postman/operation.ts
```diff
@@ -84,12 +104,11 @@ function findItem(collection: Collection, method: string, path: string): Item |
...
 
 // this func ensures that both this securityScheme and the one in HttpService share exactly the sme keys
-function findPostmanSecurityScheme(auth: RequestAuth, collection: Collection) {
+function findPostmanSecurityScheme(auth: RequestAuth, securitySchemes: PostmanSecurityScheme[]) {
   const securityScheme = transformSecurityScheme(auth, () => '1');
   if (!securityScheme) return;
 
-  const allSecuritySchemes = transformSecuritySchemes(collection);
-  return allSecuritySchemes.find(ss => isPostmanSecuritySchemeEqual(ss, securityScheme));
+  return securitySchemes.find(ss => isPostmanSecuritySchemeEqual(ss, securityScheme));
```
XVincentX: Ok so I am assuming you've decided to follow my advice of computing this somewhere else — I just need to find out where.

src/postman/operation.ts
```diff
@@ -2,24 +2,40 @@ import {
...
+  const operations: IHttpOperation[] = [];
+
+  traverseItemsAndGroups((resolvedCollection as unknown) as ItemGroup<Item>, item =>
+    operations.push(transformItem(item, securitySchemes)),
+  );
+
+  return operations;
+};
+
 export const transformPostmanCollectionOperation: PostmanCollectionHttpOperationTransformer = ({
```
XVincentX: Do we really need a function transforming a single item? I really do not see this happening and this is not even used in `transformPostmanCollectionOperations` — I guess we can probably remove this function.

src/postman/operation.ts
```diff
@@ -2,24 +2,40 @@ import {
...
 } from './transformers/securityScheme';
 import { transformServer } from './transformers/server';
 import { PostmanCollectionHttpOperationTransformer } from './types';
 import { transformDescriptionDefinition, traverseItemsAndGroups } from './util';
 
+export const transformPostmanCollectionOperations = (document: CollectionDefinition): IHttpOperation[] => {
+  const collection = new Collection(document);
+  const resolvedCollection = new Collection(collection.toObjectResolved({ variables: collection.variables }, []));
+
+  const securitySchemes = transformSecuritySchemes(collection);
```
XVincentX: Ok I like this better, transforming once forever and then passing it around 👍

src/postman/__tests__/operation.test.ts
karol-maciaszek: I think it is not. The resulting array may be longer than 4 and that is what we also want to avoid. See `expect.arrayContaining`

src/postman/__tests__/util.test.ts
karol-maciaszek: Well, I'm fixing the bug I placed there before.

Basically I assumed (and somehow wrongly verified) that `ItemGroup.forEachItem(Group)?()` does not traverse the tree recursively. I was wrong, it does. So our traverse implementation is basically useless and can be removed.

In order to make things even more complicated for the reviewer, I removed traversing in following "refactor" PR https://github.com/stoplightio/http-spec/pull/69/files#diff-4686c87b4e40c955db91ee7283af86b2L5

src/postman/operation.ts
karol-maciaszek: Yeah, this is the pattern we can efficiently use when parsing all operations.

src/postman/operation.ts
karol-maciaszek: That function is part of the `http-spec`'s core interface. Basically `http-spec` exposes two set of functions:
1. `transform%deathmetal%Operation`
2. `transform%deathmetal%Operation`

Where `%deathmetal%` is one of `Oas2`, `Oas3`, `PostmanCollection`.

https://github.com/search?q=org%3Astoplightio+transformOas2Operation&type=Code
Even in `prism` we currently* use single operation transformation.

* I refactored that here https://github.com/stoplightio/prism/compare/feat/postman-collection to use the bulk transformation function

src/postman/util.ts
karol-maciaszek: Explained here https://github.com/stoplightio/http-spec/pull/68/files#r381923071

src/postman/operation.ts
karol-maciaszek: All stays the same for single operation transformation. I just made this func more universal so it can be use sequentially. I think this comment may put some light on how `http-spec`s interface looks like: https://github.com/stoplightio/http-spec/pull/68/files#r381928417

src/postman/__tests__/operation.test.ts
XVincentX: That's true.

src/postman/util.ts
XVincentX: Aha I understand now! And I am quite sure I told you this already in a previous PR :trollface: 

Good!

src/postman/operation.ts
XVincentX: > That function is part of the http-spec's core interface


Ah ok, meh, what an usesless function anyway.

src/postman/util.ts
karol-maciaszek: :P

PR #69, opened 2020-02-12
Postman Collection - Refactor

karol-maciaszek:
- removed `traverseItemsAndGroups` in favor of using `.forEachItem`/`.forEachItemGroup` directly
- added missing tests
- removed petstore fixture - all cases are covered anyway

Screen Shot 2020-02-12 at 17 38 18
^ feeling bit sad that I can't cover those three branches, but it is cased by Postman SDK / typings bugs.

**Note2:** temporarily setting base to previous PR to show the differences


src/postman/operation.ts
```diff
@@ -20,16 +20,14 @@ import {
 } from './transformers/securityScheme';
 import { transformServer } from './transformers/server';
 import { PostmanCollectionHttpOperationTransformer } from './types';
-import { transformDescriptionDefinition, traverseItemsAndGroups } from './util';
+import { resolveCollection, transformDescriptionDefinition } from './util';
 
 export const transformPostmanCollectionOperations = (document: CollectionDefinition): IHttpOperation[] => {
-  const collection = new Collection(document);
-  const resolvedCollection = new Collection(collection.toObjectResolved({ variables: collection.variables }, []));
```
XVincentX: Good call, basically we're doing this always and it makes sense to enclose it in a function.

PR #71, opened 2020-02-16
chore: export translate parameter

mallachari:
Required by https://github.com/stoplightio/graphite/pull/348

Exports oas3 parameter translation function needed to compare path parameters between source map and virtual nodes

### DIFF
--- a/src/oas3/transformers/request.ts
+++ b/src/oas3/transformers/request.ts
@@ -28,7 +28,7 @@ function translateRequestBody(requestBodyObject: RequestBodyObject): IHttpOperat
   };
 }
 
-function translateParameterObject(parameterObject: ParameterObject): IHttpParam | any {
+export function translateParameterObject(parameterObject: ParameterObject): IHttpParam | any {
   return pickBy({
     ...omit(parameterObject, 'in', 'schema'),
     name: parameterObject.name,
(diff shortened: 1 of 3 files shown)


src/oas3/transformers/request.ts
```diff
@@ -28,7 +28,7 @@ function translateRequestBody(requestBodyObject: RequestBodyObject): IHttpOperat
   };
 }
 
-function translateParameterObject(parameterObject: ParameterObject): IHttpParam | any {
+export function translateParameterObject(parameterObject: ParameterObject): IHttpParam | any {
```
P0lip: Could we also re-export it in https://github.com/stoplightio/http-spec/blob/master/src/oas3/index.ts?
Also, what about OAS2? Is an equivalent fn already exported?

mallachari: Yes, equivalent for oas2 is already exported: https://github.com/stoplightio/http-spec/blob/master/src/oas2/transformers/params.ts#L176

We can re-export it in index as well if it's better, but it could also work with this kind of import 
`import { translateToPathParameter } from '@stoplight/http-spec/oas2/transformers/params'`

P0lip: > We can re-export it in index as well if it's better, but it could also work with this kind of import

The reason I'm suggesting re-exporting from that file is that it sort of makes it a public util, so people should be more careful about introducing changes.
If we keep it local, one will assume it's not used anywhere else.
I know it doesn't really change much from the code perspective.

Issue #53, opened 2020-01-14
Add support for server variables


lottamus:
**User story.**

As an OpenAPI v3 spec, all of my servers and server variables will be included after transforming to an IHttpService.

**Is your feature request related to a problem?**

This feature allows us to display server variables in Docs and replace server variables with their default values in @stoplight/request-maker.

**Describe the solution you'd like**

When [mapping over the http service's servers](https://github.com/stoplightio/http-spec/blob/master/src/oas3/service.ts#L38-L42), we should also include any [server variables](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#server-variable-object) as defined by the [IServer interface](https://github.com/stoplightio/types/blob/master/src/servers.ts#L8)

**Additional context**

- Add a test that ensures we are including all the server variable properties to https://github.com/stoplightio/http-spec/blob/master/src/oas3/__tests__/service.test.ts
- Need by https://github.com/stoplightio/elements/issues/196

philsturgeon:
Make sure we're looking for servers at the global level, and at the Path Item level (i.e servers can sit next to `get`, `post`, etc.) and operation level (servers can sit _inside_ `get`, `post`, etc.)

PR #76, opened 2020-02-26
feat: add server variable support

ssspear:
Related Issue: https://github.com/stoplightio/http-spec/issues/53

* Include servers and server variables when transforming an OpenAPI v3 spec to an IHttpService

### DIFF
--- a/src/oas3/__tests__/operation.test.ts
+++ b/src/oas3/__tests__/operation.test.ts
@@ -257,7 +257,6 @@ describe('transformOas3Operation', () => {
       {
         description: void 0,
         url: 'operation/server',
-        variables: void 0,
       },
     ]);
   });
@@ -356,7 +355,6 @@ describe('transformOas3Operation', () => {
       {
         description: void 0,
         url: 'path/server',
-        variables: void 0,
       },
     ]);
   });
@@ -455,7 +453,6 @@ describe('transformOas3Operation', () => {
       {
         description: void 0,
         url: 'spec/server',
-        variables: void 0,
       },
     ]);
   });
@@ -571,7 +568,7 @@ describe('transformOas3Operation', () => {
     });
   });
 
-  test('should handle server variables', () => {
+  test('should keep the server variables', () => {
     const document: Partial<OpenAPIObject> = {
       id: '?http-service-id?',
       paths: {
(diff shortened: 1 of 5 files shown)


src/oas3/transformers/servers.ts
```diff
@@ -7,13 +7,21 @@ export function translateToServers(servers: ServerObject[]): IServer[] {
...
-        description: String(value.description),
-        enum: map(value.enum, String),
-      }),
-    ),
+    variables: translateServerVariables(server.variables),
   }));
 }
+
+export function translateServerVariables(variables: DeepPartial<{ [v: string]: ServerVariableObject }> | undefined) {
+  if (!variables) return {};
```
XVincentX: The null check should be done by the caller, so let's make the parameter mandatory and let the caller figure it out

```suggestion
export function translateServerVariables(variables: DeepPartial<{ Dictionary<ServerVariableObject> }>) {
```

src/oas3/transformers/servers.ts
```diff
@@ -7,13 +7,21 @@ export function translateToServers(servers: ServerObject[]): IServer[] {
...
+  const values = mapValues<Dictionary<ServerVariableObject>, INodeVariable>(
+    pickBy(variables, isServerVariableObject),
+    value => ({
+      default: String(value.default),
+      description: value.description && String(value.description),
+      enum: value.enum && map(value.enum, String),
+    }),
+  );
+
+  return values;
```
XVincentX:
```suggestion
  return mapValues<Dictionary<ServerVariableObject>, INodeVariable>(
    pickBy(variables, isServerVariableObject),
    value => ({
      default: String(value.default),
      description: value.description && String(value.description),
      enum: value.enum && map(value.enum, String),
    }),
  );
```

src/oas3/transformers/servers.ts
XVincentX: Why is the type now a DeepPartial?

src/oas3/service.ts
```diff
@@ -35,11 +36,16 @@ export const transformOas3Service: Oas3HttpServiceTransformer = ({ document }) =
...
 
-      return {
+      const serv: IServer = {
         name: document.info?.title ?? '',
         description: server.description,
         url: server.url ?? '',
       };
+
+      const variables = translateServerVariables(server.variables);
+      if (Object.keys(variables).length) serv.variables = variables;
```
XVincentX: Why are you checking the length? Can't we set the variables object anyway?

ssspear: Yep we can. I was trying to avoid setting a property that is undefined if we didn't need to, but if that's not an issue I can change it

src/oas3/transformers/servers.ts
ssspear: I set it to DeepPartial because our document is a DeepPartial OpenApiObject, which I believe therefore makes the servers DeepPartials as well.

XVincentX: Understood.

`DeepPartial` is a a modifier that relaxes a constraint, while in general we want to put as much constraints as possible — so that the TypeChecker can prevent mistakes. In such case, I would change the definition to be `variables: Dictionary<ServerVariableObject>`

XVincentX: Ok you can ignore this — such thing is coming from the document. This is fine.

src/oas3/__tests__/operation.test.ts
```diff
@@ -570,4 +570,99 @@ describe('transformOas3Operation', () => {
       tags: [],
     });
   });
+
+  test('should handle server variables', () => {
```
XVincentX: Let's try to make this test a little bit more specific — we can change its title for example:

```suggestion
  test('should keep the server variables', () => {
```

and then instead of comparing the whole operation with an entire gigantic object we could instead make some precise assertions on `operation.server` and checking its variables

src/oas3/__tests__/operation.test.ts
```diff
@@ -455,7 +455,7 @@ describe('transformOas3Operation', () => {
       {
         description: void 0,
         url: 'spec/server',
-        variables: {},
+        variables: void 0,
```
XVincentX: You can remove this property check entirely
```suggestion
```

and the same goes for all the properties above.

src/oas3/__tests__/operation.test.ts
```diff
@@ -570,4 +570,99 @@ describe('transformOas3Operation', () => {
...
+          },
+        },
+        {
+          url: 'http://petstore.swagger.io/v2',
+          description: 'Sample Petstore Server Http',
+        },
+      ],
+    };
+
+    expect(transformOas3Operation({ document, path: '/pets', method: 'get' })).toStrictEqual({
```
XVincentX: Also, instead of testing the whole `transformOas3Operation`, we could just test the transform security function to make the test leaner. I'm ok with keeping this test though, but let's assert on the security object only

src/oas3/__tests__/operation.test.ts
```diff
@@ -570,4 +570,99 @@ describe('transformOas3Operation', () => {
...
+            },
+            port: {
+              default: '8443',
+              description: void 0,
+              enum: ['8443', '443'],
+            },
+            username: {
+              default: 'demo',
+              description: 'value is assigned by the service provider',
+              enum: void 0,
```
XVincentX: all the void 0 can be removed

src/oas3/__tests__/operation.test.ts
XVincentX: Strict equality is something really difficult in this repo/situation.

```suggestion
    expect(transformOas3Operation({ document, path: '/pets', method: 'get' })).toEqual({
```

PR #77, opened 2020-03-02
Postman Collection cookie.expires fix

karol-maciaszek:
PC typings state that `cookie.expires` is an `Date|undefined`. However an integer timestamp is also an accepted value:
- implementation: https://github.com/postmanlabs/postman-collection/blob/develop/lib/collection/cookie.js#L142
- example: https://github.com/postmanlabs/postman-collection/blob/develop/examples/collection-v2.json#L75

This PR prepares `http-spec` to handle such case.

### DIFF
--- a/src/postman/transformers/__tests__/response.test.ts
+++ b/src/postman/transformers/__tests__/response.test.ts
@@ -305,6 +305,6 @@ describe('transformResponse()', () => {
           ],
         });
       });
-    })
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


XVincentX:
@karol-maciaszek Shall we modify the postman typings as well?

karol-maciaszek:
> @karol-maciaszek Shall we modify the postman typings as well?

Definitely, I will prepare the code tomorrow.

XVincentX:
Ok I've verified and the change looks good to me. Let's bulk it together with the types change though: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42783

PR #78, opened 2020-03-03
fix: move types in dev deps

XVincentX:
Move the types as devDependencies. If you need them in your target package, you'll install them there.


marbemac:
Does this fix something? I'm not completely sure what the side effects of this are - will this break typings in consumers of `http-spec`? If so, then these should at least be in peer deps - but that doesn't really make sense to require end user to adds these manually unless there's a problem we're solving here.

XVincentX:
Not solving any specific problem, I just thought it was done by mistake. If it's on purpose (although…why?) I can close this @marbemac

package.json
```diff
@@ -31,10 +31,6 @@
     "@openapi-contrib/openapi-schema-to-json-schema": "^3.0",
     "@stoplight/json": "^3.4.0",
     "@stoplight/types": "^11.4.0",
```
karol-maciaszek: @XVincentX what about our types?

XVincentX: @karol-maciaszek What do you mean?

karol-maciaszek: I mean what about moving `"@stoplight/types": "^11.4.0",` to `devDependencies`.

XVincentX: Done!

XVincentX:
@marbemac Are you ok with this?

package.json
```diff
@@ -44,7 +39,12 @@
   },
   "devDependencies": {
     "@stoplight/scripts": "^5",
+    "@stoplight/types": "^11.4.0",
```
lottamus: This might break usage of this repo for anyone that doesn't have @stoplight/types installed since we're using const enums. Can you verify it works without installing @stoplight/types?

XVincentX: I'll move this back in the dependencies then, checking this out in all the repo is a lot of work and not worth

PR #79, opened 2020-03-03
Postman Collection: Merging IHttpOperations

### LINKED ISSUE #74: Responses with the same status code
This was discovered while testing Prism powered by Postman Collection document.

`IHttpOperation` type allows defining several `IHttpOperationResponse`’s having the same status code. But due to how Prism's negotiator works we will always consider only first response with matching code (https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/InternalHelpers.ts#L75).

For example, if we have two separate response objects with status code 200 and different mediaTypes (1. JSON, 2. XML) && we have received accept header preferring XML, then we fail to find it.

It was not possible to define several responses with the same status code in OpenAPI2/3 due to code being a property. So we never run into such a problem using OpenAPI. But with Postman Collection it is possible to define several responses within a single operation having the same codes and the rest being different. This effectively causes Prism to fail to find a proper response.

The questions:
1. Should Prism handle such cases as multiple response objects having the same status code?
2. How about the rest of our tooling, are we ready for such cases?
3. If we don't want to support ^, then maybe multiple responses from Postman Collection should be merged into one?
3a. ...ignored?
4. If we allow responses having the some code it may possible cause problem if we want to convert IHttpOperation back into OpenAPI document- do we (will )?have such case?

@XVincentX @philsturgeon @lottamus @marbemac

karol-maciaszek:
Fixes #74.

Introduces the initial capability of merging together two `IHttpOperation`'s. 

The feature is required because:
- Postman stores as separate items making it possible to e.g. define few 200 responses
- Postman may store few items (operations) defining the same endpoint

There are few things missing to call it full-blown solution:
- security merging - might be slightly required by PC, we could consider adding it
- callbacks - not required for PC

Note: The only-client of this feature currently is Postman Collection transformations

### DIFF
--- a/src/postman/operation.ts
+++ b/src/postman/operation.ts
@@ -9,7 +9,9 @@ import {
   IHttpQueryParam,
   IMediaTypeContent,
 } from '@stoplight/types';
+import { IHttpOperationResponse } from '@stoplight/types/dist';
 import { Collection, CollectionDefinition, Item, RequestAuth, Url } from 'postman-collection';
+import { mergeOperations, mergeResponses } from '../merge';
 import { transformRequest } from './transformers/request';
 import { transformResponse } from './transformers/response';
 import {
@@ -25,9 +27,11 @@ import { resolveCollection, transformDescriptionDefinition } from './util';
 export const transformPostmanCollectionOperations = (document: CollectionDefinition): IHttpOperation[] => {
   const collection = resolveCollection(document);
   const securitySchemes = transformSecuritySchemes(collection);
-  const operations: IHttpOperation[] = [];
+  let operations: IHttpOperation[] = [];
 
-  collection.forEachItem(item => operations.push(transformItem(item, securitySchemes)));
+  collection.forEachItem(item => {
+    operations = mergeOperations(operations, [transformItem(item, securitySchemes)]);
+  });
 
   return operations;
 };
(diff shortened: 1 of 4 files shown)


src/postman/operation.ts
```diff
@@ -25,9 +27,11 @@ import { resolveCollection, transformDescriptionDefinition } from './util';
 export const transformPostmanCollectionOperations = (document: CollectionDefinition): IHttpOperation[] => {
   const collection = resolveCollection(document);
   const securitySchemes = transformSecuritySchemes(collection);
-  const operations: IHttpOperation[] = [];
+  let operations: IHttpOperation[] = [];
 
-  collection.forEachItem(item => operations.push(transformItem(item, securitySchemes)));
+  collection.forEachItem(item => {
+    operations = mergeOperations(operations, [transformItem(item, securitySchemes)]);
```
karol-maciaszek: I know. However, optimizing it will make the code bit uglier :(

src/postman/operation.ts
```diff
@@ -77,7 +81,12 @@ function transformItem(item: Item, securitySchemes: PostmanSecurityScheme[]): IH
     path: getPath(item.request.url),
     summary: item.name,
     request,
-    responses: item.responses.map(transformResponse),
+    responses: item.responses
+      .all()
+      .reduce<IHttpOperationResponse[]>(
+        (merged, response) => mergeResponses(merged, [transformResponse(response)]),
+        [],
+      ) as IHttpOperation['responses'],
```
XVincentX: Why is such cast required? What are you getting back?

karol-maciaszek: I get `NonEmptyArray` in return. Actually I thought today that I could change `mergeLists` to operate on Array types. So this cast would not be needed.

XVincentX:
Let’s make it happen please.

src/postman/operation.ts
karol-maciaszek: More or less like that (I'm not sure whether this is good way to extract item type from array type).
```
function mergeLists<T>(compare: (o1: T[0], o2: T[0]) => boolean, merge: (o1: T[0], o2: T[0]) => T[0]) {
```

PR #80, opened 2020-03-04
Postman Collection: handles case when url.path is not defined

karol-maciaszek:
Fixes case when `url.path` is not defined.

There is also a pull request to `DefinitelyTyped` which fixes the types: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42833


karol-maciaszek:
Yeah, you just need to provide URL as string & without path, like: `localhost:666`

PR #81, opened 2020-03-05
Postman Collection: Required parameters

karol-maciaszek:
Makes all parameters extracted from Postman Collection required:
1) path params
2) query params
3) headers
4) object properties in generated schemas for body payload

In order to achieve 4) I replaced [to-json-schema](https://github.com/ruzicka/to-json-schema) with [json-schema-generator](https://github.com/krg7880/json-schema-generator). I'm not happy about both libs but the latter at least produces good results. The bad thing about the lib is that it looks not to be maintained. But other libs had bigger problems: not able to turn `required` on, no control over `additionalProperties` or producing JSONSchema3.

I'd say that if we want to proceed with "learning" we need some strong library and `json-schema-generator` looks like a candidate which can serve some foundation for this (uses AST inside).

More context: https://stoplight-internal.slack.com/archives/CD6HDERDX/p1583328211342900

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,10 +35,11 @@
     "@types/to-json-schema": "^0.2.0",
     "@types/type-is": "^1.6.3",
     "@types/urijs": "~1.19.4",
+    "generate-schema": "^2.6.0",
+    "json-schema-generator": "^2.0.6",
     "lodash": "^4.17.15",
     "openapi3-ts": "~1.3.0",
     "postman-collection": "^3.5.5",
-    "to-json-schema": "^0.2.3",
     "type-is": "^1.6.18",
     "urijs": "~1.19.2"
   },
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "@types/to-json-schema": "^0.2.0",
     "@types/type-is": "^1.6.3",
     "@types/urijs": "~1.19.4",
-    "generate-schema": "^2.6.0",
     "json-schema-generator": "^2.0.6",
     "lodash": "^4.17.15",
     "openapi3-ts": "~1.3.0",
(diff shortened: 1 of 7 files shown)


XVincentX:
@karol-maciaszek What do we do with this? Bring over the finish line or closing?

karol-maciaszek:
> @karol-maciaszek What do we do with this? Bring over the finish line or closing?

Ah, I realized that I need to find a better library for generating schemas out of examples. The one we're using is not capable of creating the required properties.

karol-maciaszek:
> > @karol-maciaszek What do we do with this? Bring over the finish line or closing?
> 
> Ah, I realized that I need to find a better library for generating schemas out of examples. The one we're using is not capable of creating the required properties.

XVincentX:
@karol-maciaszek Most likely all we need to do is to enable the `required` parameter in the schema generation: https://www.npmjs.com/package/to-json-schema

karol-maciaszek:
> @karol-maciaszek Most likely all we need to do is to enable the `required` parameter in the schema generation: https://www.npmjs.com/package/to-json-schema

Tested that, if I just turn the flag it will generate schema in JSONSchem3 format, which means: `required: true`. This was deprecated in following JSONSchema versions in favor of `required: string[]`.

I guess there is a converter on the market but I would like to investigate some schema generators before.

philsturgeon:
I was about to say we should avoid using this library if its this unmaintained, but looks like we used it in the past. 

image

We can move on with this so long as we accept this will need taking over. We can do that on https://github.com/openapi-contrib/ as a larger group effort if needs be, this is a generally useful tool, and I'd like to see at least one good one exist. So long as we feel like using this and forking it is generally going to be less work than sending a PR to to-json-schema adding whatever it is we need. I'll leave it up to you folks to make that call. 👍

PR #82, opened 2020-03-05
chore: improved typings for IHttpOperation merger

karol-maciaszek:
This PR improves the typings of the merger. Now the merger function is typed with array type instead of the item type. That makes it possible to pass some more complex array types, like NonEmptyArray.

There is still an open case: https://github.com/stoplightio/http-spec/pull/82/files#diff-047f6c9db2a76331a04cf84790e05f5bR85. Should we prevent creating operations with empty responses on `http-spec` level? Currently, we allow such operations which result e.g. in prism being unable to mock such request. WDYT?

### DIFF
--- a/src/postman/operation.ts
+++ b/src/postman/operation.ts
@@ -1,14 +1,4 @@
-import {
-  HttpParamStyles,
-  HttpSecurityScheme,
-  IHttpHeaderParam,
-  IHttpOperation,
-  IHttpOperationRequestBody,
-  IHttpParam,
-  IHttpPathParam,
-  IHttpQueryParam,
-  IMediaTypeContent,
-} from '@stoplight/types';
+import { HttpSecurityScheme, IHttpOperation } from '@stoplight/types';
 import { Collection, CollectionDefinition, Item, RequestAuth, Url } from 'postman-collection';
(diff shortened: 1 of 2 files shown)


XVincentX:
@karol-maciaszek What do we do with this? Bring over the finish line or close?

src/postman/operation.ts
```diff
@@ -81,12 +80,10 @@ function transformItem(item: Item, securitySchemes: PostmanSecurityScheme[]): IH
...
     request,
-    responses: item.responses
-      .all()
-      .reduce<IHttpOperationResponse[]>(
-        (merged, response) => mergeResponses(merged, [transformResponse(response)]),
-        [],
-      ) as IHttpOperation['responses'],
+    responses: item.responses.all().reduce<IHttpOperation['responses']>(
+      (merged, response) => mergeResponses(merged, [transformResponse(response)]),
+      [] as any, // @todo do we want to load operations without responses?
```
XVincentX: @karol-maciaszek Yes, I think we should load them anyway so that tooling (Prism or Platform) can show meaningful errors, such as "no responses found, but these are the one in the document:…"

karol-maciaszek: Sounds good, I'll remove the comment and let's proceed!

PR #83, opened 2020-03-06
fix: handling letter cases when merging http ops

karol-maciaszek:
Case insensitive comparison of header/mediaType/server.url when merging.

### DIFF
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -20,7 +20,7 @@ export const mergeResponses = mergeLists<IHttpOperationResponse>(
 );
 
 const mergeServers = mergeLists<IServer>(
-  (s1, s2) => s1.url.toLowerCase() === s2.url.toLowerCase(),
+  (s1, s2) => s1.url === s2.url,
   s1 => s1, // ignore server #2 is url is equal
 );
 
--- a/src/merge.ts
+++ b/src/merge.ts
@@ -1,12 +1,12 @@
 import { IHttpHeaderParam, IHttpOperation, IHttpOperationResponse, IMediaTypeContent, IServer } from '@stoplight/types';
 
 const mergeHeaders = mergeLists<IHttpHeaderParam>(
-  (h1, h2) => h1.name === h2.name,
+  (h1, h2) => h1.name.toLowerCase() === h2.name.toLowerCase(),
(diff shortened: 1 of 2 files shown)


src/merge.ts
```diff
@@ -20,12 +20,12 @@ export const mergeResponses = mergeLists<IHttpOperationResponse>(
 );
 
 const mergeServers = mergeLists<IServer>(
-  (s1, s2) => s1.url === s2.url,
+  (s1, s2) => s1.url.toLowerCase() === s2.url.toLowerCase(),
```
XVincentX: This is interesting, I do not know whether we should do this or not. urls are not case insensitive and some server do discriminate based on them. @marbemac @lottamus

XVincentX: Actually @philsturgeon you could give your idea on this too.

philsturgeon: I don’t know how many people are doing this in the wild but yeah technically they could. Let’s leave this out and see if it’s reported, either by reports we can run or by users.

karol-maciaszek: I just removed it.

I believe we should have it like that: case-insensitive host and case-sensitive path. WDYT?

XVincentX: I agree, HOSTS should be case insensitive. It this URL a full one (comprehensive of the host)?

karol-maciaszek: Well, that is even funnier. See the typings: https://github.com/stoplightio/types/blob/master/src/servers.ts#L8

We can have variables inside the server URL. I'll check this out, but it seems that we need to parse the URL in order to understand what is case-(in)?sensitive.

1) host name is case-insensitive
2) URL is case-sensitive
3) variable name: I can't find in OAS spec whether this is case sensitive or not, but I would assume it is

Now the funny part: a variable inside hostname makes it partially case-sensitive.

I guess we need to address it now or create an issue to address it later. Actually I'm also thinking about `http-spec` being responsible for server.url sanitization while transforming from different spec formats.

XVincentX: Yeah let's make a separate issue for this, we can get this over the finish line. Feel free to open an issue.

karol-maciaszek: Here is the task: https://github.com/stoplightio/http-spec/issues/84

PR #89, opened 2020-03-16
lowercase security schemes

XVincentX:
1. Security Scheme are case insensitive, so we can normalise them
2. Use a WAY better `DeepPartial` definition that does not lose typings

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "@openapi-contrib/openapi-schema-to-json-schema": "^3.0",
     "@stoplight/json": "^3.5.1",
-    "@stoplight/types": "^11.4.1",
+    "@stoplight/types": "^11.5.0",
     "@types/swagger-schema-official": "~2.0.21",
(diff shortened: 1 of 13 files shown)


marbemac:
Does it make sense to replace the `DeepPartial` in stoplight types with this one so that we use it throughout ecosystem?

XVincentX:
@marbemac 

Yes, the one that @stoplight/types is offering is broken and it gets easily into a bunch of `unknown/never` fields that makes the types incorrect and impossible to use

marbemac:
@XVincentX could you put in a PR to types that either copies the deep partial over from that other package, or re-exports it (assuming other package doesn't have too much other crap we don't need, or is only types in which case doesn't matter)?

I'll approve that and then can update this PR to use the updated deep partial, then merge.

That work?

XVincentX:
@marbemac https://github.com/stoplightio/types/pull/67

marbemac:
make sure to put correct merge message!

PR #91, opened 2020-04-23
feat: support oas2 parameter x-deprecated

mallachari:
Transforms oas2 parameter x-deprecated property.

### DIFF
--- a/src/oas2/transformers/params.ts
+++ b/src/oas2/transformers/params.ts
@@ -210,5 +210,9 @@ function buildSchemaForParameter(
     schema.minLength = 1;
   }
 
-  return { schema, description: param.description };
+  return {
+    schema,
+    description: param.description,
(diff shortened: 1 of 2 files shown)


src/oas2/transformers/params.ts
```diff
@@ -210,5 +210,9 @@ function buildSchemaForParameter(
     schema.minLength = 1;
   }
 
-  return { schema, description: param.description };
+  return {
+    schema,
+    description: param.description,
+    ...('x-deprecated' in param && { deprecated: param['x-deprecated'] }),
```
lottamus, line 216: Can you add a test for this? 😁

 https://github.com/stoplightio/http-spec/blob/master/src/oas2/transformers/__tests__/params.test.ts

mallachari: Done

PR #93, opened 2020-05-18
feat: schema examples

mallachari:
Addresses: https://github.com/stoplightio/studio/issues/381

Adds support for schema object examples (used when no media object examples are provided) to OAS2 `translateToResponses`, `translateToBodyParameter` and OAS3 `translateMEdiaTypeObject`.

OAS2:
Uses `x-examples`, `example` or `x-example` from schema object if:
- no `examples` provided in response media type object
- no `x-examples` provided in request body parameter object

OAS3:
Uses `examples` or `x-examples` from schema object if no `examples` provided in request/response media type object.
Uses `example` or `x-example` from schema object if no `examples` provided in request/response media type object.


src/oas2/schema.ts
```diff
@@ -0,0 +1,22 @@
+import { isObject } from 'lodash';
+import { Schema } from 'swagger-schema-official';
+
+export function getExamplesFromSchema(data?: Schema) {
```
XVincentX, line 4: Why is this possibility undefined? If you do not have a schema, there's no example, and the caller should know that.

```suggestion
export function getExamplesFromSchema(data: Schema) {
```

src/oas2/schema.ts
```diff
@@ -0,0 +1,22 @@
...
+  let examples;
```
XVincentX, line 5: Prefer a ternary operator to avoid the mutable reference.
```suggestion
  const examples = …
```

src/oas2/schema.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  if (isObject(data)) {
+    if ('x-examples' in data) {
+      examples = data['x-examples'];
```
XVincentX, line 9: It seems like there is a strong assumption that the data in the `x-examples` property is correct. We should validate this

src/oas2/transformers/params.ts
```diff
@@ -74,7 +75,7 @@ export function translateToHeaderParams(headers: { [headerName: string]: Header
 }
 
 export function translateToBodyParameter(body: BodyParameter, consumes: string[]): IHttpOperationRequestBody {
-  const examples = map(get(body, 'x-examples'), (value, key) => ({ key, value }));
+  const examples = map(get(body, 'x-examples') || getExamplesFromSchema(body.schema), (value, key) => ({ key, value }));
```
XVincentX, line 78: It seems to me we're handling the `x-examples` twice; once here and once in the inner function? Is this confusing?

```suggestion
  const examples = map(getExamplesFromSchema(body.schema), (value, key) => ({ key, value }));
```

src/oas3/transformers/content.ts
```diff
@@ -137,6 +142,9 @@ export function translateMediaTypeObject(mediaObject: unknown, mediaType: string
   };
 }
 
+const transformSchemaExamples = (examples: { [exampleName: string]: {} }) =>
```
XVincentX, line 145:
```suggestion
const transformSchemaExamples = (examples: Dictionary<string>: {} }) =>
```

src/oas2/schema.ts
XVincentX: Let's make the file name equal to the exported function.

src/oas3/transformers/content.ts
```diff
@@ -108,17 +108,22 @@ export function translateHeaderObject(headerObject: unknown, name: string): Opti
...
+  const jsonSchema = schema
+    ? (toJsonSchema(schema, {
+        cloneSchema: true,
+        strictMode: false,
+        keepNotSupported: ['example', 'deprecated', 'readOnly', 'writeOnly', 'xml', 'externalDocs'],
+      }) as JSONSchema4)
+    : undefined;
+
+  const example = mediaObject.example || jsonSchema?.example || jsonSchema?.['x-example'];
+  const examples = mediaObject.examples || transformSchemaExamples(jsonSchema?.examples || jsonSchema?.['x-examples']);
```
XVincentX, line 122: @marbemac @philsturgeon I really think we should NOT support or encourage `x-example` in OpenAPI 3.0. 

While I _kind_ of understand the OAS2 workaround, OAS3 has a clear solution for this and we should not invent our own methods here.

src/oas2/schema.ts
mallachari: Removed undefined type and renamed file (also moved to transformers, though it's kind of util)

src/oas2/schema.ts
mallachari: Right, I've added check if it's an object. That should be enough in that case.

src/oas2/transformers/params.ts
mallachari: First one refers to root level example (`body`), while second to schema example (`body.schema`) so those might be two different `x-examples`

src/oas3/transformers/content.ts
mallachari: I added it as it might be included even if it's not supported, but yeah, perhaps better not to encourage using something that can be done better.

marbemac: Agree - anywhere there is now an official way of accomplishing xyz in openapi 3, we should remove our custom extensions. So if there is `example` in this location, we don't need `x-example`.

src/oas3/transformers/content.ts
```diff
@@ -108,17 +108,22 @@ export function translateHeaderObject(headerObject: unknown, name: string): Opti
...
+  const jsonSchema = schema
+    ? (toJsonSchema(schema, {
+        cloneSchema: true,
+        strictMode: false,
+        keepNotSupported: ['example', 'deprecated', 'readOnly', 'writeOnly', 'xml', 'externalDocs'],
+      }) as JSONSchema4)
+    : undefined;
+
+  const example = mediaObject.example || jsonSchema?.example;
+  const examples = mediaObject.examples || transformSchemaExamples(jsonSchema?.examples || jsonSchema?.['x-examples']);
```
XVincentX, line 122: OAS3 Schema object has nor the examples nor the x-examples keywords
```suggestion
  const examples = mediaObject.examples;
```

src/oas3/transformers/content.ts
```diff
@@ -137,6 +142,8 @@ export function translateMediaTypeObject(mediaObject: unknown, mediaType: string
   };
 }
 
+const transformSchemaExamples = (examples: Dictionary<string>) => mapValues(examples, (value, key) => ({ key, value }));
```
XVincentX, line 145: This function is not used and can be removed.

src/oas2/transformers/getExamplesFromSchema.ts
```diff
@@ -0,0 +1,16 @@
+import { isObject } from 'lodash';
+import { Schema } from 'swagger-schema-official';
+
+export function getExamplesFromSchema(data: Schema) {
+  return isObject(data)
```
XVincentX, line 5: This is always true by definition (`data: Schema`) — this check is no longer necessary

src/oas2/transformers/getExamplesFromSchema.ts
```diff
@@ -0,0 +1,16 @@
...
+    ? 'x-examples' in data
+      ? isObject(data['x-examples'])
+        ? data['x-examples']
+        : void 0
+      : 'example' in data
+      ? { default: data.example }
+      : 'x-example' in data
+      ? { default: data['x-example'] }
```
XVincentX, line 13: If OAS2 supports example, we do not need to check for x-example. We just need to backfill "unsupported" parts.

src/oas3/transformers/content.ts
mallachari: Right, not OAS3, it's JsonSchema keyword and the initial @marbemac idea was to allow usage of any of those example keywords in schema object, that's why I put it here. If we can't do that we would need to stick only with `example` in that case.

philsturgeon:
@XVincentX super important context on this one: https://github.com/stoplightio/studio/issues/381#issuecomment-626944301 

It's really confusing stuff, especially as we're conflating "OAS3 Schema Object" and "JSON Schema", so lets make sure we don't turn off any keywords that "JSON Schema" users need too, like `examples` in standalone schema files.

mallachari:
> It's really confusing stuff, especially as we're conflating "OAS3 Schema Object" and "JSON Schema", so lets make sure we don't turn off any keywords that "JSON Schema" users need too, like `examples` in standalone schema files.

Alright, so let's keep JSON Schema keywords for standalone files, and for OAS operations allow nothing more than what's defined there. Though standalone schema files are not using this repo so I'll clean it up here.

src/oas2/transformers/__tests__/params.test.ts
```diff
@@ -124,6 +124,75 @@ describe('params.translator', () => {
...
+                  { key: 'example-1', value: { hello: 'world' } },
+                  { key: 'example-2', value: { foo: 'bar' } },
+                ],
+              }),
+            ]),
+          }),
+        );
+      });
+
+      it('root x-examples should take precedence over schema examples', () => {
```
XVincentX, line 163: What do you mean by precedence? As in — is the first example listed or Is the second one completely ignored?

src/oas2/transformers/__tests__/responses.test.ts
```diff
@@ -94,4 +95,75 @@ describe('responses', () => {
       expect(responses[0].contents![0].examples).toHaveLength(2);
     });
   });
+
+  describe('schema examples', () => {
+    it('given schema examples should translate to responses', () => {
```
XVincentX, line 100:
```suggestion
    it('given a response with a schema with an example', () => {
```

Also, this should be broke in multiple describe and a single `it` assertion, but this is not blocking.

src/oas2/transformers/__tests__/responses.test.ts
```diff
@@ -94,4 +95,75 @@ describe('responses', () => {
...
+                  name: 'examples value',
+                },
+              },
+            } as Schema,
+          },
+        },
+        produces,
+      );
+      expect(responses[0].contents![0]).toHaveProperty('examples', [
+        { key: 'application/json', value: { name: 'examples value' } },
```
XVincentX, line 139: I am not so sure about this, why are we ignoring the other examples?

src/oas2/transformers/__tests__/params.test.ts
mallachari: Meaning schema example ignored if there's parameter example. Though it not said explicitly in oas2 spec as there's no `examples` keyword for body parameter, I followed same approach as oas3 has which is ignoring schema example if there's one in media object: 
> if referencing a schema which contains an example, the example value SHALL override the example provided by the schema.

We may join them if that makes more sense.

src/oas2/transformers/__tests__/responses.test.ts
mallachari: That's how we did it so far for standalone JSON Schema files https://github.com/stoplightio/elements/blob/beta/src/components/Docs/HttpOperation/utils.ts#L15
Also that was the idea in initial issue:
> Note that I think we only need to choose one at a time, in this priority order: examples, x-examples, example.

Though we can always combine them if you think it's better.

XVincentX: Yes, we should combine them @mallachari — we should never lose data during the transformation.

mallachari: Alright, schema object examples are combined.

src/oas2/transformers/__tests__/params.test.ts
mallachari: How about this one? Should examples from root object (body param or response) override schema object examples (like in oas3) or rather be combined altogether?

XVincentX:
Thanks a lot!

PR #95, opened 2020-06-12
do not export postman

XVincentX:


lottamus:
@XVincentX I think this will be a breaking change for anyone importing postman from the root. I'd hate to do a major version bump just for this one change though 🤔

XVincentX:
It is effectively a fix that brings a major change. However this seems to be required to avoid requiring the whole postman library when Prism is the sole user of this at the moment. I am open to other alternatives. I would personally drop the `index.ts` file completely at this point so people will really only import what they need (we could create a `oas.ts` file or something).

src/index.ts
```diff
@@ -1,4 +1,3 @@
 export * from './oas2';
 export * from './oas3';
 export * from './oas';
-export * from './postman';
```
karol-maciaszek, line 4: Needs change here: https://github.com/stoplightio/prism/blob/master/packages/cli/src/operations.ts#L4, right?

XVincentX: Yes, this will be required in Prism

XVincentX:
@lottamus @marbemac What do you want to do with this? The issue per se has been fixed in Platform already with a aimed PR to import only OAS.

I vote to remove this index.ts file and let everybody explicitly import what they need.

marbemac:
> I vote to remove this index.ts file and let everybody explicitly import what they need.

This sounds good to me. If we do this it should be a major version bump, and would be great to do so follow-up PRs for these packages. Not sure if it's used elsewhere but a quick search didn't find anything.

- stoplight/prism-http
- analyzer
- graphite
- studio

XVincentX:
@marbemac Done

PR #97, opened 2020-06-15
fix: local references in params and responses

marbemac:
Fixes https://github.com/stoplightio/platform-internal/issues/3039
Fixes https://github.com/stoplightio/platform-internal/issues/2946

Due to new bundle strategy, incoming documents might have local $refs in them. We need to handle this case.

After these changes (compare to screenshots in 3039 linked above):

Note nice params and responses in righthand preview:

Screen Shot 2020-06-15 at 5 07 07 PM

Same thing, nice params and responses:

Screen Shot 2020-06-15 at 5 05 42 PM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,3 @@
 module.exports = {
   extends: ['@stoplight'],
-  rules: {
-    'no-param-reassign': 'off',
-  },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,6 @@
 module.exports = {
-  extends: ["@stoplight"],
+  extends: ['@stoplight'],
+  rules: {
+    'no-param-reassign': 'off',
+  },
 };
--- a/src/oas/__tests__/__snapshots__/operation.test.ts.snap
+++ b/src/oas/__tests__/__snapshots__/operation.test.ts.snap
@@ -86,7 +86,33 @@ Array [
     "id": "?http-operation-id?",
     "method": "get",
     "path": "/pets",
-    "request": Object {},
+    "request": Object {
+      "headers": Array [
+        Object {
+          "name": "Rate-Limit",
+          "schema": Object {
+            "type": "string",
+          },
+          "style": "simple",
+        },
+      ],
+      "query": Array [
+        Object {
+          "name": "skip",
+          "schema": Object {
+            "type": "string",
+          },
+          "style": "form",
+        },
+        Object {
+          "name": "limit",
+          "schema": Object {
+            "type": "string",
+          },
+          "style": "form",
+        },
+      ],
+    },
     "responses": Array [
       Object {
         "code": "200",
@@ -111,7 +137,33 @@ Array [
     "id": "?http-operation-id?",
     "method": "options",
     "path": "/pets",
-    "request": Object {},
+    "request": Object {
+      "headers": Array [
+        Object {
+          "name": "Rate-Limit",
+          "schema": Object {
+            "type": "string",
+          },
+          "style": "simple",
+        },
+      ],
+      "query": Array [
+        Object {
+          "name": "skip",
+          "schema": Object {
+            "type": "string",
+          },
+          "style": "form",
+        },
+        Object {
+          "name": "limit",
+          "schema": Object {
+            "type": "string",
+          },
+          "style": "form",
+        },
+      ],
+    },
     "responses": Array [
       Object {
         "code": "200",
@@ -209,6 +261,16 @@ Array [
         "description": "Pet object that needs to be added to the store",
         "required": true,
       },
+      "path": Array [
+        Object {
+          "name": "petId",
+          "required": true,
+          "schema": Object {
+            "type": "string",
+          },
+          "style": "simple",
+        },
+      ],
     },
     "responses": Array [
       Object {
@@ -234,15 +296,19 @@ Array [
           Object {
             "examples": Array [],
             "mediaType": "application/xml",
-            "schema": undefined,
+            "schema": Object {
+              "$ref": "#/definitions/Error",
+            },
           },
           Object {
             "examples": Array [],
             "mediaType": "application/json",
-            "schema": undefined,
+            "schema": Object {
+              "$ref": "#/definitions/Error",
+            },
           },
         ],
-        "description": undefined,
+        "description": "Our shared 404 response.",
         "headers": Array [],
       },
       Object {
@@ -373,7 +439,22 @@ Array [
       "cookie": Array [],
       "headers": Array [],
       "path": Array [],
-      "query": Array [],
+      "query": Array [
+        Object {
+          "examples": Array [],
+          "name": "skip",
+          "schema": Object {
+            "type": "string",
+          },
+        },
+        Object {
+          "examples": Array [],
+          "name": "limit",
+          "schema": Object {
+            "type": "string",
+          },
+        },
+      ],
     },
     "responses": Array [
       Object {
@@ -423,7 +504,16 @@ Array [
       },
       "cookie": Array [],
       "headers": Array [],
-      "path": Array [],
+      "path": Array [
+        Object {
+          "examples": Array [],
+          "name": "petId",
+          "required": true,
+          "schema": Object {
+            "type": "string",
+          },
+        },
+      ],
       "query": Array [],
     },
     "responses": Array [
@@ -431,6 +521,22 @@ Array [
         "code": "400",
         "contents": Array [],
         "description": "Invalid ID supplied",
+        "headers": Array [],
+      },
+      Object {
+        "code": "404",
+        "contents": Array [
+          Object {
+            "encodings": Array [],
+            "examples": Array [],
+            "mediaType": "*/*",
+            "schema": Object {
+              "$ref": "#/components/schemas/Error",
+              "$schema": "http://json-schema.org/draft-04/schema#",
+            },
+          },
+        ],
+        "description": "Our shared 404 response.",
         "headers": Array [],
       },
       Object {
--- a/src/oas/__tests__/__snapshots__/operation.test.ts.snap
+++ b/src/oas/__tests__/__snapshots__/operation.test.ts.snap
@@ -1,0 +1,480 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`oas operation openapi v2 1`] = `
+Array [
+  Object {
+    "description": "",
+    "id": "?http-operation-id?",
+    "iid": "addPet",
+    "method": "post",
+    "path": "/pets",
+    "request": Object {
+      "body": Object {
+        "contents": Array [
+          Object {
+            "examples": Array [],
+            "mediaType": "application/json",
+            "schema": Object {
+              "$ref": "#/definitions/Pet",
+            },
+          },
+          Object {
+            "examples": Array [],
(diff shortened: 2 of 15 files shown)


src/oas/__tests__/__snapshots__/operation.test.ts.snap
```diff
@@ -0,0 +1,586 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`oas operation openapi v2 1`] = `
```
marbemac, line 3: I opted for a operation snapshot test because it's easy and efficient to verify resolve results.

.eslintrc.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
-  extends: ["@stoplight"],
+  extends: ['@stoplight'],
+  rules: {
+    'no-param-reassign': 'off',
+  },
```
XVincentX, line 5: Why such rule?

src/oas/__tests__/__snapshots__/operation.test.ts.snap
XVincentX: On the other hand, I see this object is just gigantic and there's no way somebody is going to review this line by line. I suggest to make few and focused assertions rather than drop this here.

src/utils.ts
```diff
@@ -54,3 +54,15 @@ export function URI(url: string | URI = '') {
 
 export const isDictionary = (maybeDictionary: unknown): maybeDictionary is Dictionary<unknown> =>
   isObjectLike(maybeDictionary);
+
+export const getLocalRefValue = (document: unknown, $ref: string): unknown => {
+  return get(document, pointerToPath($ref));
+};
+
+export const maybeResolveLocalRef = (document: unknown, target: unknown): unknown => {
```
XVincentX, line 62: 1. I suggest to keep the name `resolveLocalRef`, since there's not a real external condition driving this and the early return feels more an optimisation rather than a real resolving condition
2. Since we're passing the target, we can be kind of confident that the returned object will still be of the target type — so I suggest this:
```suggestion
export const maybeResolveLocalRef = <T>(document: unknown, target: T): T => {
```

src/oas/accessors.ts
```diff
@@ -1,12 +1,18 @@
...
 
 export function getValidOasParameters<ParamType extends { name: string; in: string }>(
+  document: unknown,
   operationParameters: ParamType[] | undefined,
   pathParameters: ParamType[] | undefined,
 ) {
-  return unionBy(operationParameters, pathParameters, (parameter?: ParamType) =>
-    isObject(parameter) ? `${parameter.name}-${parameter.in}` : 'invalid',
-  ).filter(isObject);
+  const op = map(operationParameters, x => maybeResolveLocalRef(document, x) as ParamType);
```
XVincentX, line 10:
```suggestion
  const resolvedOperationParameters = map(operationParameters, x => maybeResolveLocalRef(document, x) as ParamType);
```

src/oas/accessors.ts
```diff
@@ -1,12 +1,18 @@
...
+  const pp = map(pathParameters, x => maybeResolveLocalRef(document, x) as ParamType);
```
XVincentX, line 11: Same with this

src/oas2/guards.ts
```diff
@@ -13,3 +13,10 @@ export const isTagObject = (maybeTagObject: unknown): maybeTagObject is Tag => {
 
   return false;
 };
+
+export const isResponseObject = (maybeResponseObject: unknown): maybeResponseObject is Response =>
+  isObject(maybeResponseObject) &&
+  ('description' in maybeResponseObject ||
+    'schema' in maybeResponseObject ||
+    'headers' in maybeResponseObject ||
+    'examples' in maybeResponseObject);
```
XVincentX, line 22: Isn't this a little bit too relaxed?

- description should be a string
- headers, headers and examples should be objects

src/oas2/transformers/responses.ts
```diff
@@ -1,51 +1,72 @@
-import { IHttpOperationResponse } from '@stoplight/types';
+import { DeepPartial, Dictionary, IHttpOperationResponse, Optional } from '@stoplight/types';
```
XVincentX, line 1:
```suggestion
import type { DeepPartial, Dictionary, IHttpOperationResponse, Optional } from '@stoplight/types';
```

src/oas2/transformers/responses.ts
```diff
@@ -1,51 +1,72 @@
...
-  }));
-
-  const translatedResponses = {
-    code: statusCode,
-    description: response.description,
-    headers,
-    contents,
-  };
+function responseTransformer(document: DeepPartial<Spec>) {
+  return function translateToResponse(
```
XVincentX, line 12: Ok two observations

1. I'll keep saying that the dependencies go at the end. I still propose to reverse the order.
2. In my opinion it does not really make that much sense to just apply this pattern in a single place. Either we do (1) and we do it everywhere — or we just don't do it at all and put the `document` as a parameter.

src/oas2/transformers/responses.ts
```diff
@@ -1,51 +1,72 @@
...
+    produces: string[],
+    response: unknown,
```
XVincentX, line 14: Why is this now unknown if then you're casting it to `Response` on line 30?

src/oas2/transformers/responses.ts
```diff
@@ -1,51 +1,72 @@
...
+    }
+
+    return translatedResponses;
+  };
 }
 
 export function translateToResponses(
-  responses: { [name: string]: Response },
+  document: DeepPartial<Spec>,
+  responses: unknown,
```
XVincentX, line 59: Why did you change this to `unknown`?

src/oas2/guards.ts
marbemac: There were no guards before - I pulled over the guard from OAS3 and tweaked for OAS2. I guess it's a matter of wether we want to do a full validation? Yes description should be a string (although null is fine as well), headers should be an object.. but then should we further validate the object properties, etc? Kind of a can of worms.

src/oas2/transformers/responses.ts
marbemac: This allowed for nice composition with the `map` where it's used - undoing this will make the map more cumbersome - maybe it's worth it just for that?

src/oas2/transformers/responses.ts
marbemac: This was REALLY weird - if you pull this down you'll find that line 30 is the only place in this function that, for whatever reason, still thinks response is `unknown` even after the guard higher up in the function. All other use of `response` in this function correctly identifies that it's a `Response`. Any ideas?

src/oas2/transformers/responses.ts
marbemac: Because it is unknown - we do the guard check later. It's similar to the oas3 transformer - I guess just had not been improved here yet?

src/oas/__tests__/__snapshots__/operation.test.ts.snap
marbemac: 🤷‍♂️ I went through it line by line and found it quite useful to verify all the right things are happening.

src/oas2/transformers/responses.ts
XVincentX: TypeScript is correct.

Although the arrow function has the context from outside, there's really no guarantee that the `isResponseObject` will be called *before* the arrow function. They are still separate entities. To fix this:

```ts
    const resolvedResponse = maybeResolveLocalRef(document, response);
    if (!isResponseObject(resolvedResponse)) return;

    const headers = translateToHeaderParams(resolvedResponse.headers || {});
    const objectifiedExamples = chain(
      resolvedResponse.examples || (resolvedResponse.schema ? getExamplesFromSchema(resolvedResponse.schema) : void 0),
    )
      .mapValues((value, key) => ({ key, value }))
      .values()
      .value();

    const contents = produces.map(produceElement => ({
      mediaType: produceElement,
      schema: resolvedResponse.schema as JSONSchema4,
      examples: objectifiedExamples.filter(example => example.key === produceElement),
    }));
```

```diff
- response = maybeResolveLocalRef(document, response);
+ const resolvedResponse = maybeResolveLocalRef(document, response);
```

src/oas2/transformers/responses.ts
XVincentX: Ah yeah ok — that makes sense.

src/utils.ts
marbemac: Actually I don't think this is a good idea. The $ref can be pointing at anything - best to keep it as unknown, and do a specific guard check after attempting the resolve (which is what we do) to verify the final type.

.eslintrc.js
marbemac: For this - https://github.com/stoplightio/http-spec/pull/97/files#diff-c84e068125096414207643df176f27fbR17

I suppose I could ignore at the line level if you prefer?

src/oas2/transformers/responses.ts
XVincentX: If I've localised the line you're referring to correctly, I do not see any improvement in the composition; I rather see a partial application which, in case we want to pursue that approach, I'd extend to the maximum extent:

1. All arguments in one line:

```ts
function responseTransformer(document: DeepPartial<Spec>, produces: string[], response: unknown, statusCode: string): Optional<IHttpOperationResponse>
```

2. Partial application for the first two arguments:

```ts
  return compact<IHttpOperationResponse>(
    map<Dictionary<unknown>, Optional<IHttpOperationResponse>>(
      responses,
      partial(responseTransformer, document, produces),
    ),
  );
```

.eslintrc.js
XVincentX: I cannot identify the line

src/oas2/transformers/responses.ts
marbemac: Oh right the oas2 one uses partial - I was thinking of the oas3 version - https://github.com/stoplightio/http-spec/blob/fix/local-references/src/oas3/transformers/responses.ts#L43.

I don't really care that much, happy to change. Given oas3 version, do you still think we should remove this pattern from here (and the oas3 spot)? Again, I don't mind either way.

src/utils.ts
XVincentX: Hmm I am afraid I am not following:

```ts
type Test = {
    name: string,
    age: number,
    $ref?: string
}

declare function resolve<T>(target: T): Omit<T,'$ref'> & {[k:string]:unknown}

const t: Test = {
    name: 'Vincenzo',
    age: 31
};

const resolved = resolve(t);
```

Isn't the resolved *at least* a T?

src/oas2/transformers/responses.ts
XVincentX: I'd use partial everywhere — it's more appropriate. We can evaluate dependency injection at a separate stage, but that (as you noticed) will require a through discussion.

src/utils.ts
marbemac: No - what if the $ref points to something that does not exist, or to some other random thing. It's why we do the specific guard check after the maybe resolve - https://github.com/stoplightio/http-spec/blob/fix/local-references/src/oas3/transformers/responses.ts#L18-L19.

XVincentX: That's fair. Better be defensive — as long it's not `any` I'm ok with it.

PR #100, opened 2020-06-30
fix(oas): openapi-schema-to-json-schema may throw

P0lip:
long story short - http-spec should never throw an error, otherwise X task in Y project won't succeed and ultimately no preview panel will be rendered for a given document or node.

This is occurring quite often: https://sentry.io/organizations/stoplightio/issues/1673691344/?project=1549141&query=is%3Aunresolved+properties&statsPeriod=14d

### DIFF
--- a/src/oas3/transformers/content.ts
+++ b/src/oas3/transformers/content.ts
@@ -110,13 +110,19 @@ export function translateMediaTypeObject(mediaObject: unknown, mediaType: string
 
   const { schema, encoding } = mediaObject;
 
-  const jsonSchema = schema
-    ? (toJsonSchema(schema, {
+  let jsonSchema: Optional<JSONSchema4>;
+
+  if (isObject(schema)) {
+    try {
+      jsonSchema = toJsonSchema(schema, {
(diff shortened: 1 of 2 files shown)


src/oas3/transformers/content.ts
```diff
@@ -110,13 +110,19 @@ export function translateMediaTypeObject(mediaObject: unknown, mediaType: string
...
+    try {
+      jsonSchema = toJsonSchema(schema, {
         cloneSchema: true,
         strictMode: false,
         keepNotSupported: ['example', 'deprecated', 'readOnly', 'writeOnly', 'xml', 'externalDocs'],
-      }) as JSONSchema4)
-    : undefined;
+      }) as JSONSchema4;
+    } catch {
+      // happens
```
P0lip, line 123: shall we report it somehow so we have a clue that XYZ thing in the convert fails and we might need to do something about it?

XVincentX: We talked about this for ages. There's no way to log stuff from HTTP Spec, so we'll just ignore it :(

PR #109, opened 2020-07-20
fix: more bulletproof maybeResolveLocalRef

P0lip:
Fixes https://github.com/stoplightio/platform-internal/issues/3589 + one potential issue with circular refs.

### DIFF
--- a/src/__tests__/utils.test.ts
+++ b/src/__tests__/utils.test.ts
@@ -120,9 +120,7 @@ describe('URI()', () => {
 
 describe('maybeResolveLocalRef()', () => {
   it('handles invalid $refs', () => {
-    expect(maybeResolveLocalRef({ a: true }, { $ref: '#a' })).toStrictEqual({
-      $ref: '#a',
-    });
+    expect(maybeResolveLocalRef({ a: true }, { $ref: '#a' })).toBeNull();
   });
 
   it('handles circular references', () => {
--- a/src/__tests__/utils.test.ts
+++ b/src/__tests__/utils.test.ts
@@ -1,6 +1,6 @@
 import * as urijs from 'urijs';
 
-import { URI } from '../utils';
+import { maybeResolveLocalRef, URI } from '../utils';
 
 describe('URI()', () => {
   it('instantiates from string', () => {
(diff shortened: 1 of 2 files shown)


src/__tests__/utils.test.ts
```diff
@@ -117,3 +117,25 @@ describe('URI()', () => {
     });
   });
 });
+
+describe('maybeResolveLocalRef()', () => {
+  it('handles invalid $refs', () => {
+    expect(maybeResolveLocalRef({ a: true }, { $ref: '#a' })).toStrictEqual({
+      $ref: '#a',
```
P0lip, line 124: @XVincentX is this okay from Prism perspective or not necessarily?

XVincentX: My point will always be the same: this should halt the parsing of this specific section and send an error up to the caller.

This will effectively let people feed Prism with an invalid document which will lead to an error on its own.

P0lip: > My point will always be the same: this should halt the parsing of this specific section and send an error up to the caller.

Yeah, I know your point of view and totally understand it.
This is not okay from the perspective of Studio, unfortunately. :disappointed: 

Would returning null/undefined be better here?

XVincentX: None of these will make me happy — but with null we'd be at least ignoring the whole thing for real and that's better than leaving the reference unresolved. I'd vote for that.

P0lip: > None of these will make me happy 

Yeah, totally aware of it, that only errors will satisfy Prism, but it is what it is :man_shrugging: 
To be frank, I'm really leaning towards diverging http-spec, not sure whether it makes share the codebase.
The Studio and Prism needs vary a lot, and sharing the same package makes everything only more complicated for each tool respectively.

Speaking of code - will return null instead then.

XVincentX: > To be frank, I'm really leaning towards diverging http-spec, not sure whether it makes share the codebase.

To me it does not. I've proposed that in the past and I received a no a response. Maybe if we join the forces, we can win this time?

src/utils.ts
```diff
@@ -55,13 +55,21 @@ export function URI(url: string | URI = '') {
...
 
 export const maybeResolveLocalRef = (document: unknown, target: unknown): unknown => {
   if (hasRef(target) && isLocalRef(target.$ref)) {
-    return getLocalRefValue(document, target.$ref);
+    try {
+      return getLocalRefValue(document, target.$ref);
+    } catch (ex) {
+      if (ex instanceof URIError) {
+        return target;
+      }
```
XVincentX, line 69: As agreed, let's return null.

src/__tests__/utils.test.ts
P0lip: > Maybe if we join the forces, we can win this time?

:crossed_fingers:

PR #112, opened 2020-08-11
fix: filter out foreign scopes

XVincentX:
This PR will make sure that only the scopes that are defined in the operation security scope will be effectively copied in the final operation.

I am open to better alternatives to implement the filter.

### DIFF
--- a/src/oas2/service.ts
+++ b/src/oas2/service.ts
@@ -1,5 +1,5 @@
-import { HttpSecurityScheme, IHttpService, IOauth2Flow, IServer } from '@stoplight/types';
-import { compact, filter, flatMap, isString, keys, mapValues, pickBy } from 'lodash';
+import { HttpSecurityScheme, IHttpService, IServer } from '@stoplight/types';
+import { compact, filter, flatMap, isString, keys, pickBy } from 'lodash';
 
 import { Oas2HttpServiceTransformer } from '../oas/types';
 import { isTagObject } from './guards';
@@ -60,16 +60,22 @@ export const transformOas2Service: Oas2HttpServiceTransformer = ({ document }) =
       return keys(sec).map(key => {
         const ss = securitySchemes.find(securityScheme => securityScheme.key === key);
         if (ss && ss.type === 'oauth2') {
-          return {
-            ...ss,
-            flows: mapValues(ss.flows, (flow: IOauth2Flow) => ({
+          const flows = {};
+          for (const flowKey in ss.flows) {
+            const flow = ss.flows[flowKey];
+            flows[flowKey] = {
               ...flow,
               scopes: pickBy(flow.scopes, (_val: string, scopeKey: string) => {
                 const secKey = sec[key];
                 if (secKey) return secKey.includes(scopeKey);
-                return false;
+                return undefined;
               }),
-            })),
+            };
+          }
+
+          return {
+            ...ss,
+            flows,
           };
         }
 
@@ -79,7 +85,6 @@ export const transformOas2Service: Oas2HttpServiceTransformer = ({ document }) =
   );
 
   if (security.length) {
-    //@ts-ignore I hate doing this, but unfortunately Lodash types are (rightfully) very loose and put undefined when it can't happen
     httpService.security = security;
   }
 
--- a/src/oas2/service.ts
+++ b/src/oas2/service.ts
@@ -1,5 +1,5 @@
-import { HttpSecurityScheme, IHttpService, IServer } from '@stoplight/types';
-import { compact, filter, flatMap, isString, keys } from 'lodash';
+import { HttpSecurityScheme, IHttpService, IOauth2Flow, IServer } from '@stoplight/types';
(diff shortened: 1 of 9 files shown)


src/oas3/accessors.ts
```diff
@@ -12,15 +12,34 @@ export function getSecurities(
...
 
   return opSchemesPairs.map(opSchemePair =>
     opSchemePair
       .map(opScheme => {
-        return { ...definitions[opScheme], key: opScheme };
+        const definition = definitions[opScheme];
+
+        if (isSecurityScheme(definition) && definition.type === 'oauth2') {
+          // Put back only the flows that are part of the current definition
+          definition.flows = mapValues(definition.flows, (flow: OAuthFlowObject) => ({
```
P0lip, line 36: Hmm, aren't you mutating the original document here?

XVincentX: Ah! Good point, I'm afraid I am. I'll fix that!

XVincentX: Done in 9a4c8b4

lottamus:
@XVincentX do we need to do the same thing for global security in the service? 

- https://github.com/stoplightio/http-spec/blob/f2f0acb3ca2957223aff2f805781fe202df314dc/src/oas2/service.ts#L68
- https://github.com/stoplightio/http-spec/blob/f2f0acb3ca2957223aff2f805781fe202df314dc/src/oas3/service.ts#L77

XVincentX:
> @XVincentX do we need to do the same thing for global security in the service?
 

The global securities do not need to be filtered — the filtering is only required locally, where we need to grab from the global filter what's needed.

lottamus:
> The global securities do not need to be filtered — the filtering is only required locally, where we need to grab from the global filter what's needed.

By global security, I'm not talking about security schemes definition. I'm talking about the security requirement object: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#securityRequirementObject

image


```yaml
openapi: 3.0.0
info:
  title: My API
  version: '1.0'
servers:
  - url: 'http://localhost:3000'
paths: {}
components:
  schemas: {}
  securitySchemes:
    API Key - 1:
      type: oauth2
      description: ''
      flows:
        implicit:
          authorizationUrl: ''
          refreshUrl: ''
          scopes:
            scope_1: ''
            scope_2: ''
security:
  - API Key - 1:
      - scope_1

```

XVincentX:
You're right. Filters added.

src/oas3/service.ts
```diff
@@ -67,13 +69,28 @@ export const transformOas3Service: Oas3HttpServiceTransformer = ({ document }) =
...
+          };
+        }
+
+        return ss;
       });
     }),
   );
+
   if (security.length) {
+    //@ts-ignore
```
XVincentX, line 93: I hate doing this, but Lodash types are effectively very loose and would require other checks that we do not really need.

src/oas3/service.ts
```diff
@@ -67,13 +69,28 @@ export const transformOas3Service: Oas3HttpServiceTransformer = ({ document }) =
...
+          };
+        }
+
+        return ss;
       });
     }),
   );
+
   if (security.length) {
+    //@ts-ignore I hate doing this, but unfortunately Lodash types are (rightfully) very loose and put undefined when it can't happen
```
lottamus, line 93: Instead of ts-ignore, why not ditch the lodash mapValues and use a for in loop? I tried it locally and it removes the typing issue

```ts
          const flows = {};
          for (const key in ss.flows) {
            const flow = ss.flows[key];
            flows[key] = {
              ...flow,
              scopes: pickBy(flow.scopes, (_val: string, scopeKey: string) => {
                const secKey = sec[key];
                if (secKey) return secKey.includes(scopeKey);
                return undefined;
              }),
            };
          }
```

src/oas2/service.ts
```diff
@@ -58,13 +58,28 @@ export const transformOas2Service: Oas2HttpServiceTransformer = ({ document }) =
...
+          };
+        }
+
+        return ss;
       });
     }),
   );
+
   if (security.length) {
+    //@ts-ignore I hate doing this, but unfortunately Lodash types are (rightfully) very loose and put undefined when it can't happen
```
lottamus, line 82: 🤮

src/oas2/service.ts
```diff
@@ -58,13 +58,28 @@ export const transformOas2Service: Oas2HttpServiceTransformer = ({ document }) =
...
+              scopes: pickBy(flow.scopes, (_val: string, scopeKey: string) => {
+                const secKey = sec[key];
+                if (secKey) return secKey.includes(scopeKey);
+                return false;
+              }),
+            })),
+          };
+        }
+
+        return ss;
```
lottamus, line 76: Is this the same block of code that's in the oas3 service? Could we break it out to a common function to use in both spots?

src/oas3/service.ts
XVincentX: That is a great idea. I'll do the change now.

PR #113, opened 2020-08-14
fix: correctly iterate on the key

XVincentX:
I misunderstood the signature of `pickBy`, writing incorrect data. 🤦

### DIFF
--- a/src/oas3/accessors.ts
+++ b/src/oas3/accessors.ts
@@ -37,7 +37,7 @@ export function getSecurities(
             ...definition,
             flows: mapValues(definition.flows, (flow: OAuthFlowObject) => ({
               ...flow,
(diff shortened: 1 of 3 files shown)


src/oas3/accessors.ts
```diff
@@ -37,7 +37,7 @@ export function getSecurities(
             ...definition,
             flows: mapValues(definition.flows, (flow: OAuthFlowObject) => ({
               ...flow,
-              scopes: pickBy(flow.scopes, (key: string) => flattenPairs[opScheme].includes(key)),
+              scopes: pickBy(flow.scopes, (_val: string, key: string) => flattenPairs[opScheme].includes(key)),
```
lottamus, line 40: Ah it's backwards from what I'd expect too

PR #118, opened 2020-09-07
fix(oas): security flows transformation is incorrect

P0lip:
We should really stop using `DeepPartial<Xyz>` pattern - this is incorrect for most of the time.
We usually don't validate the provided input in any way before operating on it.
Moreover, using `pickBy` is just not quite great IMHO when it comes to typings, same as `get` is.

`Object.assign` was removed as it clobbered typing and was tad inconsistent with the how oas2 was implemented.

### DIFF
--- a/src/oas2/guards.ts
+++ b/src/oas2/guards.ts
@@ -1,9 +1,10 @@
-import { Dictionary } from '@stoplight/types';
 import { isObject } from 'lodash';
 import type { Response, Security, Tag } from 'swagger-schema-official';
 
+import { isDictionary } from '../utils';
+
 export function isSecurityScheme(maybeSecurityScheme: unknown): maybeSecurityScheme is Security {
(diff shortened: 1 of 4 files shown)


src/oas2/guards.ts
```diff
@@ -1,9 +1,10 @@
-import { Dictionary } from '@stoplight/types';
 import { isObject } from 'lodash';
 import type { Response, Security, Tag } from 'swagger-schema-official';
 
+import { isDictionary } from '../utils';
+
 export function isSecurityScheme(maybeSecurityScheme: unknown): maybeSecurityScheme is Security {
-  return isObject(maybeSecurityScheme) && typeof (maybeSecurityScheme as Dictionary<unknown>).type === 'string';
```
XVincentX, line 6: 👍

XVincentX (approved):

XVincentX:
This project should just be written in JavaScript.

PR #125, opened 2021-02-02
fix(oas3): resolve $refs in media contents

P0lip:
Needed for https://github.com/stoplightio/platform-internal/issues/5218


domagojk (approved):
Code looks good and I can confirm it works in studio

lottamus (approved):

PR #126, opened 2021-02-07
fix: maybeResolveLocalRef keeps broken and follows $refs 

P0lip:

### DIFF
--- a/src/__tests__/utils.test.ts
+++ b/src/__tests__/utils.test.ts
@@ -119,6 +119,22 @@ describe('URI()', () => {
 });
 
 describe('maybeResolveLocalRef()', () => {
+  it('follows $refs', () => {
+    expect(
+      maybeResolveLocalRef(
+        {
+          a: {
+            $ref: '#/b',
+          },
+          b: {
+            c: 'woo',
+          },
+        },
+        { $ref: '#/a/c' },
+      ),
+    ).toEqual('woo');
+  });
+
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -55,16 +55,12 @@ export function URI(url: string | URI = '') {
...
-
 export const maybeResolveLocalRef = (document: unknown, target: unknown): unknown => {
-  if (hasRef(target) && isLocalRef(target.$ref)) {
+  if (isDictionary(document) && hasRef(target) && isLocalRef(target.$ref)) {
     try {
-      return getLocalRefValue(document, target.$ref);
+      return resolveInlineRef(document, target.$ref);
     } catch {
-      return null;
+      return target;
```
P0lip, line 63: Here is why I decided to do it
https://github.com/stoplightio/platform-internal/pull/5264/files#r570212260
Feels like having retaining broken $refs may actually provide some value.

src/__tests__/utils.test.ts
```diff
@@ -119,8 +119,24 @@ describe('URI()', () => {
 });
 
 describe('maybeResolveLocalRef()', () => {
+  it('follows $refs', () => {
```
P0lip, line 122: This passes now thanks to `resolveInlineRef`

marbemac (approved):
I like it 👍

PR #127, opened 2021-02-11
feat: sort out $ref resolving once and for all

P0lip:
When signing off https://github.com/stoplightio/platform-internal/issues/5218 I spotted there is at least one additional place I didn't cover... This freaked me out a little bit, and I decided to revamp our approach to $ref resolving.
I'm uncertain whether we'll merge it or not, but opening a PR to have a forum to discuss the pros and cons.


src/oas/__tests__/__snapshots__/operation.test.ts.snap
```diff
@@ -15,14 +15,110 @@ Array [
             "examples": Array [],
             "mediaType": "application/json",
             "schema": Object {
-              "$ref": "#/definitions/Pet",
+              "properties": Object {
```
marbemac, line 18: Hmm does this mean that when we serialize and send the operation up to our backend to store in the database, that it'll end up getting stored fully dereferenced? 🤔

lottamus: We likely only want to resolve local references that are not nested within a "schema" object.

P0lip: > Hmm does this mean that when we serialize and send the operation up to our backend to store in the database, that it'll end up getting stored fully dereferenced? thinking

Yeah, but we're doing the same now after all.


> We likely only want to resolve local references that are not nested within a "schema" object.

hmmm, yeah, you might be right.

marcelltoth:
@P0lip Fun fact, we are actually doing something very similar locally on the elements level: https://github.com/stoplightio/elements/blob/main/packages/elements-core/src/utils/ref-resolving/resolvedObject.ts

Not to say anything about its applicability for the greater ecosystem but it's been working fairly reliably for us, for what we use it for. It enables us to finally write and wire-up $ref-ignorant components while still handling all the resolving.

P0lip:
We have bundling now, so gonna to close this one.

PR #128, opened 2021-02-15
feat(deps): drop @openapi-contrib/openapi-schema-to-json-schema

P0lip:
Needed for stoplightio/platform-internal#5359
The code I left is what we actually use from that library, with a few changes.

### DIFF
--- a/src/oas3/transformers/schema.ts
+++ b/src/oas3/transformers/schema.ts
@@ -2,6 +2,7 @@ import type { JSONSchema4, JSONSchema4TypeName } from 'json-schema';
 import { isObject } from 'lodash';
 import type { SchemaObject } from 'openapi3-ts/src/model/OpenApi';
 
+// based on https://github.com/openapi-contrib/openapi-schema-to-json-schema
 const settings = {
   MIN_INT_32: 0 - Math.pow(2, 31),
   MAX_INT_32: Math.pow(2, 31) - 1,
--- a/src/oas3/transformers/schema.ts
+++ b/src/oas3/transformers/schema.ts
@@ -51,7 +51,7 @@ function isJsonSchema4TypeName(maybeJSONSchema4TypeName: string): maybeJSONSchem
 }
 
 // Convert from OpenAPI 3.0 `SchemaObject` to JSON schema v4
-export function convertFromSchema(schema: SchemaObject, options: { pruneNotSupported: string[] }): JSONSchema4 {
+export function translateSchemaObject(schema: SchemaObject, options: { pruneNotSupported: string[] }): JSONSchema4 {
   const clonedSchema = convertSchema(schema, {
     _notSupported: options.pruneNotSupported,
     _structs: ['allOf', 'anyOf', 'oneOf', 'not', 'items', 'additionalProperties'],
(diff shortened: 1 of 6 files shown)


src/oas3/transformers/schema.ts
```diff
@@ -0,0 +1,152 @@
+import type { JSONSchema4, JSONSchema4TypeName } from 'json-schema';
```
marbemac, line 1: Interesting that this converter is in the `oas3` transformers directory. Does this process not apply to `oas2`?

IF this is indeed rather generic "turn json schema into openapi compliant draft 4" functionality, maybe it belongs in the `oas` directory?

marbemac (changes requested):
A few requests: 

Could you summarize the changes from original library in this PR so that we have a record of it?

Building on the other comment I made, should we expose this to consumers of http-spec so that other areas in our ecosystem that were using `openapi-contrib/openapi-schema-to-json-schema` can instead use this version that's included as part of the http-spec library?

And follow up to that, which other areas in our ecosystem are using `openapi-contrib/openapi-schema-to-json-schema` that might be able to now use this? :)

src/oas3/transformers/schema.ts
P0lip: > Interesting that this converter is in the oas3 transformers directory. Does this process not apply to oas2?

perhaps it's because oas2 doesn't support draft 4? not sure.

P0lip:
> Could you summarize the changes from original library in this PR so that we have a record of it?

yup, will do!

> Building on the other comment I made, should we expose this to consumers of http-spec so that other areas in our ecosystem that were using openapi-contrib/openapi-schema-to-json-schema can instead use this version that's included as part of the http-spec library?

:+1: that sounds good.

> And follow up to that, which other areas in our ecosystem are using openapi-contrib/openapi-schema-to-json-schema that might be able to now use this? :)

it's only http-spec that uses that library

marbemac (approved):
> it's only http-spec that uses that library

Ahh ok well in that case sorry, let's not expose it until we need to 😊 (so, revert the export).

Otherwise looks good!

src/oas3/transformers/schema.ts
philsturgeon: OAS2 and OAS3 are both based on draft 4 (technically oas3 is draft 5 but they’re basically the same thing in any way that matters).

P0lip: ah, sorry, I wasn't clear.
What I meant was that the oas2 json schema flavor has different keywords, that the library doesn't recognize.

I might be wrong, though, and perhaps the library yields decent results for OAS2.

philsturgeon:
The library is not intended for use with OAS2 but using it on an OAS2 library shouldn’t blow anything up too badly as the schema object just got more keywords in 3 compared to 2, no breaking changes just more stuff. 

Anyway, it’s a _little_ concerning that we’ve ditched the library due to problems because it just means the library is going to stagnate, but if we can get to OS3.1 quickly we might not have to worry about that. 

--
Phil Sturgeon
DevRel @ Stoplight.io

PR #134, opened 2021-03-05
fix: server scheme validation

mallachari:
Fixes https://github.com/stoplightio/elements/issues/904

Validates server schemes against invalid characters (i.e. whitespaces or leading numbers) that would throw if passed to `URI`.

### DIFF
--- a/src/oas2/guards.ts
+++ b/src/oas2/guards.ts
@@ -21,3 +21,7 @@ export const isResponseObject = (maybeResponseObject: unknown): maybeResponseObj
     'schema' in maybeResponseObject ||
     'headers' in maybeResponseObject ||
     'examples' in maybeResponseObject);
+
+export function isValidScheme(scheme: unknown) {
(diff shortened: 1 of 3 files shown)


src/oas2/guards.ts
```diff
@@ -21,3 +21,7 @@ export const isResponseObject = (maybeResponseObject: unknown): maybeResponseObj
     'schema' in maybeResponseObject ||
     'headers' in maybeResponseObject ||
     'examples' in maybeResponseObject);
+
+export function isValidScheme(scheme: unknown) {
+  return typeof scheme === 'string' && /^[a-zA-Z]\w+$/.test(scheme);
```
lottamus, line 26: Why don't we do a hard check against the possible values?

image

mallachari: Good point, that would make it even safer

marcelltoth (approved):
Looks perfect from an @stoplightio/undefined perspective. 👏

marbemac (approved):
👏

PR #136, opened 2021-04-01
feat(oas3): support all we need for now except schema

P0lip:
Addresses stoplightio/platform-internal#5776 

Needs https://github.com/stoplightio/types/pull/79

I didn't add tests for the following

> ensure that spaceDelimited and pipeDelimited are allowed on object type (rc0 Extended Functionality)

This is rather out of scope for `http-spec`. I'd say it's more of a Studio work, to make sure it's capable of generating a good example, i.e. http://spec.openapis.org/oas/v3.1.0#style-examples.
`http-spec` does not care what kind of type is given, it only cares about the style, and this is preserved.

> ensure that style, explode and allowReserved are allowed on multipart/form-data (rc0 Extended Functionality)

Reasoning: we don't seem to have such a need. This is how `IMediaTypeContent` looks like

```ts
export interface IHttpContent {
  schema?: JSONSchema4 | JSONSchema6 | JSONSchema7;
  examples?: (INodeExample | INodeExternalExample)[];
  encodings?: IHttpEncoding[];
}

export interface IMediaTypeContent extends IHttpContent {
  mediaType: string;
}
```

OAS 3.x.x already supports style, explode and allowReserved on `application/x-www-form-urlencoded`, but we ignore it, so I'm skeptical about adding it at this point, as we clearly do not need it yet.

See http://spec.openapis.org/oas/v3.0.0#fixed-fields-12

### DIFF
--- a/src/oas/operation.ts
+++ b/src/oas/operation.ts
@@ -5,25 +5,29 @@ import { Spec } from 'swagger-schema-official';
 
 import { HttpOperationTransformer } from '../types';
 
-const methods = ['get', 'post', 'put', 'delete', 'options', 'head', 'patch', 'trace'];
+const DEFAULT_METHODS = ['get', 'post', 'put', 'delete', 'options', 'head', 'patch', 'trace'];
 
 export function transformOasOperations(
   document: Spec | OpenAPIObject,
   transformer: HttpOperationTransformer<any>,
+  methods: string[] | null = DEFAULT_METHODS,
 ): IHttpOperation[] {
   const paths = keys(get(document, 'paths'));
 
   return flatten(
-    map(paths, path =>
-      keys(get(document, ['paths', path]))
-        .filter(pathKey => methods.includes(pathKey))
-        .map(method =>
-          transformer({
-            document,
-            path,
-            method,
-          }),
-        ),
-    ),
+    map(paths, path => {
+      let operations = keys(get(document, ['paths', path]));
+      if (methods !== null) {
+        operations = operations.filter(pathKey => methods.includes(pathKey));
+      }
+
+      return operations.map(method =>
+        transformer({
+          document,
+          path,
+          method,
+        }),
+      );
+    }),
   );
 }
--- a/src/oas3/__tests__/service.test.ts
+++ b/src/oas3/__tests__/service.test.ts
@@ -209,7 +209,7 @@ describe('oas3 service', () => {
     });
   });
 
-  it('should filter out out scopes', () => {
+  it('should filter out scopes', () => {
     const document: Partial<OpenAPIObject> = {
       openapi: '3.0.0',
       components: {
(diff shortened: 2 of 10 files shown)


src/oas/operation.ts
```diff
@@ -5,25 +5,29 @@ import { Spec } from 'swagger-schema-official';
 
 import { HttpOperationTransformer } from '../types';
 
-const methods = ['get', 'post', 'put', 'delete', 'options', 'head', 'patch', 'trace'];
+const DEFAULT_METHODS = ['get', 'post', 'put', 'delete', 'options', 'head', 'patch', 'trace'];
 
 export function transformOasOperations(
   document: Spec | OpenAPIObject,
   transformer: HttpOperationTransformer<any>,
+  methods: string[] | null = DEFAULT_METHODS,
```
domagojk, line 13: Is there a place where `null` is passed for `methods` ? I don't see any 🤷‍♂️ 

Seems to me this can be defined without it: 
```suggestion
  methods: string[] = DEFAULT_METHODS,
```

src/oas3/__tests__/service.test.ts
```diff
@@ -209,7 +209,7 @@ describe('oas3 service', () => {
     });
   });
 
-  it('filters out scopes', () => {
+  it('should filter out out scopes', () => {
```
domagojk, line 212:
```suggestion
  it('should filter out scopes', () => {
```

src/oas3/operation.ts
```diff
@@ -20,11 +20,11 @@ export function transformOas3Operations(document: OpenAPIObject): IHttpOperation
 
 export const transformOas3Operation: Oas3HttpOperationTransformer = ({ document, path, method }) => {
   const pathObj = maybeResolveLocalRef(document, get(document, ['paths', path])) as PathsObject;
-  if (!pathObj) {
+  if (typeof pathObj !== 'object' || pathObj === null) {
```
domagojk, line 23: just curious, `!pathObj` is not enough? Could that variable be a string or?

domagojk (changes requested):
A few nitpicks but it looks good overall 👍

src/oas/operation.ts
P0lip: Going to be useful in graphite, not in this repo.

src/oas3/operation.ts
P0lip: yeah, can be anything

domagojk (approved):
I would comment the `null` thing with what exactly it will be useful since it's not clear in the code. But I can live without it as well so I'm approving it 🙂 

Reasons for not handling `multipart/form-data` props you mentioned seem good to me, but best if @wmhilton or @marbemac also confirms it.

marbemac (approved):

PR #137, opened 2021-04-12
feat(oas3): support requestBody references

rainum:
Addresses https://github.com/stoplightio/platform-internal/issues/5189

This PR adds support for references in the `requestBody` of operation object with corresponding tests. Also it fixes some snapshots where this feature was dropped intentionally.

According to [OAS3 Specification](http://spec.openapis.org/oas/v3.0.0#fixed-fields-7) operation object's `requestBody` can be a [Reference Object](http://spec.openapis.org/oas/v3.0.0#referenceObject).

### DIFF
--- a/src/oas3/transformers/request.ts
+++ b/src/oas3/transformers/request.ts
@@ -14,6 +14,7 @@ import type {
 import { compact, map, omit, partial, pickBy } from 'lodash';
 import type { OpenAPIObject, ParameterObject, RequestBodyObject } from 'openapi3-ts';
 
+import { isDictionary, maybeResolveLocalRef } from '../../utils';
 import { translateMediaTypeObject } from './content';
 
 function translateRequestBody(
@@ -69,11 +70,16 @@ export function translateToRequest(
     params[key].push(translateParameterObject(parameter));
   }
 
-  const body = requestBodyObject ? translateRequestBody(document, requestBodyObject) : { contents: [] };
+  let body;
+  if (isDictionary(requestBodyObject)) {
+    const resolvedRequestBodyObject = maybeResolveLocalRef(document, requestBodyObject) as RequestBodyObject;
+    body = translateRequestBody(document, resolvedRequestBodyObject);
+  } else {
+    body = { contents: [] };
+  }
 
   return {
     body,
-
     headers: params.header,
     query: params.query,
     cookie: params.cookie,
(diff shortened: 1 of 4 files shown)


src/oas3/transformers/request.ts
```diff
@@ -69,11 +70,16 @@ export function translateToRequest(
     params[key].push(translateParameterObject(parameter));
   }
 
-  const body = requestBodyObject ? translateRequestBody(document, requestBodyObject) : { contents: [] };
+  let body;
+  if (isDictionary(requestBodyObject)) {
+    const resolvedRequestBodyObject = maybeResolveLocalRef(document, requestBodyObject) as RequestBodyObject;
```
P0lip, line 75: This type assertion is rather unsafe.
Unfortunately for us, `http-spec` needs to gracefully handles as many errors as possible, so we need to account for having an incorrect value.
image
```suggestion
    const resolvedRequestBodyObject = maybeResolveLocalRef(document, requestBodyObject);
```

src/oas3/transformers/request.ts
```diff
@@ -69,11 +70,16 @@ export function translateToRequest(
...
+    body = translateRequestBody(document, resolvedRequestBodyObject);
```
P0lip, line 76:
```suggestion
    body = isRequestBodyObject(resolvedRequestBodyObject) ? translateRequestBody(document, resolvedRequestBodyObject) : { contents: [] };
```

P0lip: We will need to add such a guard to `src/oas3/guards.ts`.

```ts
export const isRequestBodyObject = (maybeRequestBodyObject: unknown): maybeRequestBodyObject is RequestBodyObject =>
  isDictionary(maybeRequestBodyObject) && isDictionary(maybeRequestBodyObject.content) // alternatively we could check for ContentObject here, but it's good enough for now;
```

P0lip (commented):
Looks good, though we need to get rid of that type assertion.

src/oas3/transformers/request.ts
rainum: Good point @P0lip, thank for your suggestions! I applied all of them and added an additional test to make sure it wont fail.

rainum:
Comments are addressed, PR is ready for re-review.

P0lip (approved):

lottamus (approved):

PR #138, opened 2021-04-15
feat(oas): more robust schema object converter

P0lip:
The first part of https://github.com/stoplightio/platform-internal/issues/5879

I added a bit more than listed in the issue above, as there were a few wins to make here and there.


src/oas/transformers/schema/__tests__/schema.spec.ts
```diff
@@ -0,0 +1,112 @@
+import { translateSchemaObject } from '..';
+import { OASSchemaObject } from '../types';
+
+const translate = (schemaObject: OASSchemaObject) => translateSchemaObject(schemaObject, { pruneNotSupported: [] });
+
+describe('translateSchemaObject', () => {
```
P0lip, line 6: @philsturgeon I went a bit beyond the initial requirements we had in the ticket and applied some suggestions you left there.
It's not the newest draft, but it's also not Draft 6. :laughing: 
Would you like to take a look to see if there's something else you find useful covering in this converter?

@marbemac you may also want to look into this at your leisure.

src/oas/transformers/schema/__tests__/schema.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    it('should translate x-example', () => {
+      expect(
+        translate({
+          type: 'string',
+          'x-example': 'Cat',
+        }),
+      ).toStrictEqual({
+        $schema: 'http://json-schema.org/draft-07/schema#',
+        type: 'string',
+        'x-example': 'Cat', // kept for backwards compatibility, we may remove it in future (or maybe in this PR, let's figure it out!)
```
marbemac, line 67: How much work to get rid of it?

P0lip: I'm mostly concerned about JSE here.
It doesn't support arrayish examples, so we'd need to think of some UI for it.

OAS 3.1 says `example` is still valid, but `examples` is encouraged. 
`examples` is also more valid from JSON Schema's point of view.

philsturgeon (commented):
Draft 7 is a brilliant compromise! Thank you for doing this awesome work.

src/oas/transformers/schema/__tests__/schema.spec.ts
P0lip: > AFAIK it doesn't support arrayish examples, so we'd need to think of some UI for it.

Reusing input we have for `enum` might be a good trade-off.
Are you good with it @marbemac?

P0lip:
@marbemac should we consider this as a breaking change?
I'd be leaning towards that.
In such case, how do you feel about creating a beta branch and releasing this PR as beta?
There will be more related PRs coming up, and we could release them at once.

src/oas/transformers/schema/__tests__/schema.spec.ts
marbemac: Hmm well the trick is that examples can occur anywhere right? Even on properties that are objects, which would not work well ui-wise w the simple tag input we use for enum 🤔.

marbemac:
> @marbemac should we consider this as a breaking change?
> I'd be leaning towards that.
> In such case, how do you feel about creating a beta branch and releasing this PR as beta?
> There will be more related PRs coming up, and we could release them at once.

So long as it's not open forever, I've got no problem w a beta release :).

src/oas/transformers/schema/__tests__/schema.spec.ts
P0lip: > Hmm well the trick is that examples can occur anywhere right? 

yeah, same as enum, and we don't support complex `enum`.

> Even on properties that are objects, which would not work well ui-wise w the simple tag input we use for enum.

Our current `example` is a plain text field, so from my point of view the change would be rather subtle

> which would not work well ui-wise w the simple tag input we use for enum thinking.

I agree, but as said, both example and enum fields are already limited to primitive values.

We can keep the existing `example` input if you wish, and simply do the conversion (`schema.examples = [example]`) later on, thoughts?

philsturgeon: If we are trying to keep the scope of the example > examples change to feature parity with what we currently have, then @P0lip is entirely right.

image

We jut want to do this, and I believe the enum UI covers that nicely.

If we want to add "make an example for a whole object" then we can do that as a follow up feature, right?

src/oas/transformers/schema/__tests__/schema.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+      $schema: 'http://json-schema.org/draft-07/schema#',
+      type: 'object',
+      properties: {
+        test: {
+          $id: 'test',
+        },
+      },
+    });
+  });
+
```
domagojk, line 27: Just in case, I would consider adding a test to make sure passed schema is really not mutated.

```suggestion
it('should not mutate schema', () => {
    let before: OASSchemaObject = {
      type: 'object',
      properties: {
        test: {
          id: 'test',
        },
      },
    };

    translate(before);

    expect(before).toStrictEqual({
      type: 'object',
      properties: {
        test: {
          id: 'test',
        },
      },
    });
  });
```

domagojk (changes requested):
It looks good! I've only suggested adding one more test to check mutability.

domagojk (approved):

PR #139, opened 2021-04-18
feat(oas): use schema converter where appropriate 

P0lip:
Needs https://github.com/stoplightio/http-spec/pull/138 and https://github.com/stoplightio/types/pull/80


domagojk (approved):
LGTM 👍 

I would maybe consider making `translateSchemaObject` accept `undefined`, and return `undefined` in that case, so you don't have to check `isDictionary(schema)` before using it, but I'm ok either way.

PR #141, opened 2021-04-21
feat: support jsonSchemaDialect

P0lip:
For https://github.com/stoplightio/platform-internal/issues/5792
I'm going to easy this converter in Spectral.


paulatulis (approved):
Honestly @P0lip you understand this way better than I do, and the tests all pass. I'm approving and have faith that if there's a glaring issue, it's something we can roll back and address as necessary.

PR #149, opened 2021-05-18
fix: bump stoplight json

pytlesk4:


wmhilton (approved):

wmhilton (commented):
You could bump it to 3.14 now actually. That'd be cool, bc it's the pi edition

PR #163, opened 2021-10-29
fix: dereferencing response headers objects before response assembling

radzserg:
Fixes the issue when a [header is set as $ref](https://spec.openapis.org/oas/v3.1.0#response-object)

```
 headers: {
            'X-Page': {
              $ref: '#/components/headers/xPage',
            },
          },
```

In this case, we need to dereference the header object first and then assemble response object.

### DIFF
--- a/src/oas3/transformers/__tests__/responses.test.ts
+++ b/src/oas3/transformers/__tests__/responses.test.ts
@@ -175,4 +175,58 @@ describe('translateToOas3Responses', () => {
       },
     ]);
   });
+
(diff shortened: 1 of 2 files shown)


marbemac (approved):
looks good to me!

P0lip (commented):
Looks good.

What about OAS2 tho? Left on purpose?
Having a quick peek at OAS2 it appears to me we may suffer from a similar issue there, but at the same time, OAS2 doesn't allow reusable headers, so well, we might be good 🤔

PR #165, opened 2021-11-26
fix: no content means no content

mpodlasin:
Related to https://github.com/stoplightio/elements/issues/1074

It aligns the behaviour for oas2 and oas3. 

Previously there was an awkward situation where, for example, 204 No Content response seemed to... have content, for example `application/json`. Schema was empty, but non-empty `contents` array still strongly suggests some content present.

I decided to go full on. If users didn't provide schema *or* examples in oas2, it means that the response has no contents.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@
 *.log*
 *-debug.log*
 *-error.log*
+
+# Vim buffer files
+*.swp
--- a/src/oas2/__tests__/operation.test.ts
+++ b/src/oas2/__tests__/operation.test.ts
@@ -71,6 +71,39 @@ describe('transformOas2Operation', () => {
     ).toMatchSnapshot();
   });
 
+  it('should translate operation with no response body', () => {
+    const document: Partial<Spec> = {
+      swagger: '2.0',
+      info: {
+        title: 'title',
+        version: '1.0',
+      },
+      produces: ['application/json'],
+      paths: {
+        '/users/{userId}': {
+          delete: {
+            operationId: 'oid',
+            description: 'odesc',
+            summary: 'osum',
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -32,3 +32,6 @@
 *.log*
 *-debug.log*
 *-error.log*
+
+# Vim buffer files
```
mallachari, line 36: I thought you're done with vim 😉

mallachari (approved):
I'm good with the changes. Though I think my approval won't count here

src/oas3/__tests__/operation.test.ts
```diff
@@ -29,6 +29,37 @@ describe('transformOas3Operation', () => {
...
+      },
+    };
+
+    expect(
+      transformOas3Operation({
+        path: '/users/{userId}',
+        method: 'delete',
+        document,
+      }),
+    ).toMatchSnapshot();
```
P0lip, line 60: let's not use snapshots for such a test

src/oas2/__tests__/operation.test.ts
```diff
@@ -71,6 +71,39 @@ describe('transformOas2Operation', () => {
...
+      },
+    };
+
+    expect(
+      transformOas2Operation({
+        path: '/users/{userId}',
+        method: 'delete',
+        document,
+      }),
+    ).toMatchSnapshot();
```
P0lip, line 104: ditto

src/oas3/__tests__/operation.test.ts
mpodlasin: Done, Mister.

lottamus (approved):

PR #166, opened 2021-12-21
feat: http operation extensions

mallachari:
Addresses: https://github.com/stoplightio/platform-internal/issues/7772
~Requires: https://github.com/stoplightio/types/pull/97~

Transforming operation `x-extensions` properties to `HttpOperation` `extensions` property.

### DIFF
--- a/src/oas/accessors.ts
+++ b/src/oas/accessors.ts
@@ -1,5 +1,5 @@
 import { Extensions } from '@stoplight/types';
-import { isObject, map, unionBy } from 'lodash';
+import { fromPairs, isObject, map, unionBy } from 'lodash';
 
 import { maybeResolveLocalRef } from '../utils';
 
@@ -30,7 +30,5 @@ export function getOasTags(tags: unknown): string[] {
 }
 
 export function getExtensions(target: Record<string, unknown>): Extensions {
-  return Object.fromEntries(
-    Object.entries(target).filter(([key]) => key.startsWith('x-') && !ROOT_EXTENSIONS.includes(key)),
-  );
+  return fromPairs(Object.entries(target).filter(([key]) => key.startsWith('x-') && !ROOT_EXTENSIONS.includes(key)));
 }
--- a/src/oas/accessors.ts
+++ b/src/oas/accessors.ts
@@ -5,6 +5,8 @@ import { maybeResolveLocalRef } from '../utils';
 
 type ParamTypeBase = { name: string; in: string };
 
+const ROOT_EXTENSIONS = ['x-internal'];
+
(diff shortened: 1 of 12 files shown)


src/oas/accessors.ts
```diff
@@ -25,3 +26,15 @@ const isValidOasParameter = (parameter: Partial<ParamTypeBase>): parameter is Pa
 export function getOasTags(tags: unknown): string[] {
   return Array.isArray(tags) ? tags.filter(tag => typeof tag !== 'object').map(String) : [];
 }
+
+export function getExtensions(target: object): Extensions {
```
P0lip, line 30:
```suggestion
export function getExtensions(target: Record<string, unknown>): Extensions {
```

src/oas/accessors.ts
```diff
@@ -25,3 +26,15 @@ const isValidOasParameter = (parameter: Partial<ParamTypeBase>): parameter is Pa
...
+  return Object.keys(target)
+    .filter(key => key.startsWith('x-'))
+    .reduce(
+      (obj, key) => ({
+        ...obj,
+        [key]: target[key],
+      }),
+      {},
+    );
+}
```
P0lip, line 40:
```suggestion
  return Object.fromEntries(Object.entries(target).filter(([key]) => key.startsWith('x-'));
}
```

P0lip: I think we may want to exclude extensions we already consume, i.e. `x-internal`. Currently, it's added twice, once as `-internal` and the other time within `extensions`

mallachari: Yeah, I was aware of `x-internal` and thought we might still want to keep it in the object to not have exceptions. But can surely exclude it.

mallachari:
@stoplightio/plaid could you take a look? We need your approval here 😉

src/oas/accessors.ts
```diff
@@ -25,3 +28,7 @@ const isValidOasParameter = (parameter: Partial<ParamTypeBase>): parameter is Pa
 export function getOasTags(tags: unknown): string[] {
   return Array.isArray(tags) ? tags.filter(tag => typeof tag !== 'object').map(String) : [];
 }
+
+export function getExtensions(target: Record<string, unknown>): Extensions {
+  return fromPairs(Object.entries(target).filter(([key]) => key.startsWith('x-') && !ROOT_EXTENSIONS.includes(key)));
```
Nezteb, line 33: Why are we excluding `x-internal`?

mallachari: It is already being included separately as `internal` property in `HttpOperation`, so we don't want to double it

marbemac (approved):

P0lip (approved):

PR #167, opened 2022-01-13
feat: drop urijs

P0lip:
2 reasons for doing so:
- enabling `esModuleInterop` in https://github.com/stoplightio/http-spec/pull/168 without changing too much of the code
- we just don't need that dependency at all


lottamus (approved):
Looks good assuming it doesn't break anything downstream (which I assume it does not)

P0lip:
It shouldn't, yeah.

PR #170, opened 2022-01-25
fix: create security item for each scope

domagojk:
For stoplightio/platform-internal#9530

It makes sure security is created for each scope. I've removed mapping to keys and then looking in `flattenPairs` in `getSecurities` function for it to work. Why was this done in the first place, I don't know. Maybe there is some use case I'm not aware of? 
All tests are passing tho 🤷‍♂️

### DIFF
--- a/src/oas3/__tests__/accessors.test.ts
+++ b/src/oas3/__tests__/accessors.test.ts
@@ -113,4 +113,62 @@ describe('getOas3Securities', () => {
       ),
     ).toStrictEqual([[]]);
   });
+
+  it('should return security for each scope', () => {
+    const res = getSecurities(
+      {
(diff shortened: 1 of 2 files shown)


wmhilton:
Code seems reasonable. Much more readable than the old code TBH. Got a screenshot of it yalced into platform-internal showing it fixes the issue?

marbemac (approved):
This looks solid to me 👍.

domagojk:
> Got a screenshot of it yalced into platform-internal showing it fixes the issue?

You can check it in elements. So this is the screenshot of a yalced version of the openapi spec provided in issue:

Screenshot 2022-01-27 at 11 25 36

PR #175, opened 2022-03-16
feat: setup package exports

P0lip:
BREAKING CHANGE: defined package exports
BREAKING CHANGE: Node >=14.13 is required


P0lip:
The intent here is to release a beta version to avoid any nasty implications as http-spec is used in a lot of spots, so it's safer to update each package as needed.

package.json
```diff
@@ -12,12 +12,36 @@
...
     "**/*"
   ],
   "engines": {
-    "node": ">=10"
+    "node": "^12.20 || >=14.13"
+  },
+  "type": "commonjs",
+  "main": "./index.js",
+  "types": "./index.d.ts",
+  "exports": {
```
P0lip, line 24: alternatively, if we reallllly want to avoid breaking changes, we'd need to define each entry based on our usage in platform, prism, and everywhere else, but I believe these defaults are good. 
We should be really importing anything very deeply nested

package.json
```diff
@@ -12,12 +12,36 @@
     "url": "https://github.com/stoplightio/http-spec"
   },
   "license": "Apache-2.0",
-  "main": "src/index.ts",
   "files": [
     "**/*"
   ],
   "engines": {
-    "node": ">=10"
+    "node": "^12.20 || >=14.13"
```
lottamus, line 19: Should we take this opportunity to drop node 12? 🤓

P0lip: why not :D
```suggestion
    "node": ">=14.13"
```

src/oas/index.ts
```diff
@@ -1 +1,2 @@
 export * from './operation';
+export { translateSchemaObject } from './transformers/index';
```
lottamus, line 2: Do we also need to export `translateLogo`?

lottamus (approved):

src/oas/index.ts
P0lip: nope, not used publicly

PR #176, opened 2022-03-24
refactor(oas): rework all translators

P0lip:
Here's a high-level overview of changes:
- getting rid of lodash to make sure types aren't clobbered
- sorted out broken typings in many spots
  Do note that this goes beyond updating types, but actually involves changes in implementation
  
  Examples:
    - schema located under header objects weren't translated at all
    - header parameter could be assigned any style despite `IHttpHeaderParam` declaring style to be "simple".
 
  There's many more, but that can usually be figured based on the change.
- slicing translate functions into smaller pieces which will be needed by the stable IDs PRs
- unified a way values are picked up:
   -  `a ?? b`,
   - `typeof a === 'x' ? a : 'some deal'`
   - `String(a ?? '')`
   etc.

  Everything is now either explicitly casted to the desired data type (if the field is considered to be an important one), or we pick a default if the value is invalid. If optional fields such as description are missing or invalid, we drop them using `...pickBy(obj, guard)`
- got rid of `...omit` notation to ensure we don't end up having properties we don't care about 
- increased the reliance on `unknown`, got rid of most of outstanding `any`


src/context.ts
```diff
@@ -0,0 +1,9 @@
+import { TransformerContext } from './types';
+import { maybeResolveLocalRef } from './utils';
+
+export function createContext<T extends Record<string, unknown>>(document: T): TransformerContext<T> {
```
P0lip, line 4: this will take `generateId` later on

src/types.ts
```diff
@@ -1,15 +1,29 @@
...
 
 export type HttpOperationTransformer<T> = (opts: T) => IHttpOperation;
+
+export type TransformerContext<T extends Fragment = Fragment> = {
+  maybeResolveLocalRef(target: unknown): unknown;
+  document: T;
+};
+
+export type TranslateFunction<T extends Fragment, P extends unknown[], R extends unknown = unknown> = (
+  this: TransformerContext<T>,
```
P0lip, line 25: `this` plays quite nicely with array functions etc.

One can do

```
headers.map(translateHeader, this);
```

etc.

src/oas/__tests__/__snapshots__/operation.test.ts.snap
```diff
@@ -3,6 +3,7 @@
 exports[`oas operation openapi v2 1`] = `
 Array [
   Object {
+    "deprecated": false,
```
marbemac, line 6: Kind of feel that these opt-in props, where the default is false (deprecated, internal, required, explode, etc) - perhaps best to just leave them off rather than assign this default value of false? Basically only add them when they're true, to keep the doc size and general "noise" down 🤔. Or maybe it doesn't matter 🤷‍♂️

src/oas/__tests__/__snapshots__/operation.test.ts.snap
```diff
@@ -516,14 +569,12 @@ Array [
     ],
     "servers": Array [
       Object {
-        "description": undefined,
+        "name": "Swagger Petstore",
```
marbemac, line 572: I forget what the logic is here - why does the server name end up getting defaulted to the name of the service again? Was there an issue or rationale back in the day that led to the behavior 🤔

src/oas/__tests__/__snapshots__/service.test.ts.snap
```diff
@@ -37,23 +37,17 @@ Object {
...
     Object {
-      "description": undefined,
       "name": "Swagger Petstore",
       "url": "http://petstore.swagger.io/v2",
     },
   ],
   "tags": Array [
     Object {
       "description": "Everything about your Pets",
-      "externalDocs": Object {
```
marbemac, line 53: Do we not care about this anymore?

src/oas3/transformers/__tests__/__snapshots__/responses.test.ts.snap
```diff
@@ -48,9 +46,13 @@ Array [
             "value": 1000,
           },
         ],
+        "explode": false,
         "name": "fake-header-name-1",
         "schema": Object {
+          "$schema": "http://json-schema.org/draft-07/schema#",
           "format": "int32",
+          "maximum": 2147483647,
+          "minimum": -2147483648,
```
marbemac, line 55: 🤔 ?

src/oas3/transformers/__tests__/__snapshots__/request.test.ts.snap
```diff
@@ -4,7 +4,7 @@ exports[`translateOas3ToRequest give a request body should translate it 1`] = `
 Object {
   "body": Object {
     "contents": Array [
-      "translateMediaTypeObject({}, {\\"schema\\":{\\"deprecated\\":true}}, content-a, {\\"content-a\\":{\\"schema\\":{\\"deprecated\\":true}}})",
+      "translateMediaTypeObject([\\"content-a\\",{\\"schema\\":{\\"deprecated\\":true}}], 0, [[\\"content-a\\",{\\"schema\\":{\\"deprecated\\":true}}]])",
```
marbemac, line 7: This doesn't look quite right.

marbemac (changes requested):

src/oas3/transformers/__tests__/__snapshots__/responses.test.ts.snap
P0lip: long story short - was a bug to clobbered typings. We've never "translated" schemas located under headers in responses. These values are inserted by the aforementioned converter. 
`int32` format is specific to OAS, so we got to convert the values if we want to be compliant with JSON Schema and have `exclusiveMinimum` etc. handled properly.
https://github.com/stoplightio/http-spec/blob/master/src/oas/transformers/schema/keywords/format.ts#L24

src/oas/__tests__/__snapshots__/service.test.ts.snap
P0lip: It's not exposed on https://github.com/stoplightio/types/blob/master/src/graph.ts#L10, so I don't think it's needed.
The reason it was added in the first place is that we used to do something like https://github.com/stoplightio/http-spec/blob/master/src/oas3/service.ts#L115L118
In other words, we took the entire contents rather than cherry-picking what we actually want.
This isn't an isolated case, there were other scenarios where we did something like `omit(schema, keyA, keyB)` etc., which yet again isn't ideal because we may end up having plenty of junk if one has some, say, x-extensions in their document.
I went with allow-listing approach here rather than ignore-listing.

src/oas/__tests__/__snapshots__/operation.test.ts.snap
P0lip: Yeah, I was torn when it comes to truly boolean-ish values like deprecated or internal.
I mostly wanted to avoid inserting invalid values like numbers, strings, etc.
As things stand, we do the following:

```
internal: object['x-internal'] // or deprecated, etc.
```
which is awful since there's no guarantee that that value will be authored as a boolean in the document.
They should be booleans for most of the time, particularly when one uses Studio, but you never know. YAML + code editor... :D



> where the default is false
with `explode` and some others I don't recall now (probably style, so not a boolean) it's actually not that obvious, as the default tends to differ based on some other value.
https://github.com/stoplightio/http-spec/blob/d0be4880b74047e5a3b0ee859a7f10483217433a/src/oas3/transformers/content.ts#L49
image

I forgot to include that in the list of changes, but I also considered that lack of normalization as a bug, because I don't think we handled that elsewhere in our tooling.

marbemac: But basically we're storing the "default" value as if it's the user's choice. The "lack of a value" is a specific choice, basically "i want to use the default behavior". By always storing a concrete value, we lose the ability to understand if the user wants/wanted to use the default, or they specifically wanted a particular value. 🤔

src/oas3/transformers/__tests__/__snapshots__/responses.test.ts.snap
marbemac: Hmm, not sure this is practical though... with this, everybody is going to see these random maximum/minimum values in their docs, etc.

src/oas3/transformers/__tests__/__snapshots__/request.test.ts.snap
P0lip: it's correct because these are array params. That's what's given when you do `entries(media).map(translateMedia)`.
To be fair, we should probably don't mock that at all and simply print the value.
I don't really think there's any real benefit of having that small function mocked here.

src/oas/__tests__/__snapshots__/operation.test.ts.snap
P0lip: hmmmm, that is correct, although the reality is that our end tooling doesn't really figure out these defaults. 
It's usually "take what http-spec gives us" kind of a situation, so the more we do it in http-spec, the better the support is.

I'm torn, though. Feels like we could actually leave these values unset if not explicitly authored.
If authored, we should pick only when it's a valid value, i.e. boolean.
What do you think?

src/oas/__tests__/__snapshots__/operation.test.ts.snap
marbemac: Meh, can leave for now I suppose - I'm also not sure why we do it for global servers

src/oas3/transformers/__tests__/__snapshots__/request.test.ts.snap
P0lip: Got rid of that mock in https://github.com/stoplightio/http-spec/pull/176/commits/2fca01b189917fd66e3f23843ccc50480a8f022d

marbemac (approved):

PR #177, opened 2022-03-28
feat: generate stable ids

P0lip:


src/oas3/transformers/servers.ts
```diff
@@ -1,38 +1,41 @@
+import { pickBy } from '@oclif/parser/lib/util';
```
marbemac, line 1: Prob shouldn't pull from oclif right?

P0lip: ah yeah, true, dumb auto-completion

marbemac (approved):
let's give this beast a whirl

PR #178, opened 2022-04-06
feat(oas): use fnv32 for models and http operations

P0lip:


src/oas2/transformers/__tests__/params.test.ts
```diff
@@ -70,35 +69,50 @@ describe('params.translator', () => {
...
           },
-        ),
-      ).toMatchSnapshot();
+        },
+      );
+
+      expect(params).toHaveLength(1);
+      expect(params[0]).toMatchSnapshot({
+        schema: {
+          'x-stoplight-id': expect.any(String),
```
marbemac, line 85: Because if not provided it gets assigned a random id or something?

P0lip: no, it's not a random id, but I feel like there's no need to test the id here. We have some in other spots that verify the whole HTTP service/operation output.

marbemac (approved):

PR #179, opened 2022-04-07
fix(oas2): redundant ids

P0lip:
I spotted this accidentally.
It's not a huge issue as these ids wouldn't be consumed either way, but it's still good to address it to avoid any confusion in the future.

I also removed the remaining `jest.mock`

### DIFF
--- a/src/oas2/transformers/__tests__/request.test.ts
+++ b/src/oas2/transformers/__tests__/request.test.ts
@@ -7,81 +7,170 @@ import {
 } from 'swagger-schema-official';
 
 import { createContext } from '../../../oas/context';
-import {
-  translateFromFormDataParameters,
-  translateToBodyParameter,
-  translateToHeaderParam,
-  translateToPathParameter,
-  translateToQueryParameter,
-} from '../params';
 import { translateToRequest as _translateToRequest } from '../request';
 
-jest.mock('../params');
-
-const translateToRequest = (path: Record<string, unknown>, parameters: any[]) => {
-  const ctx = createContext({ consumes: ['*'], paths: { '/api': { parameters } } });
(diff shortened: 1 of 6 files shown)


src/oas2/transformers/__tests__/request.test.ts
```diff
@@ -7,81 +7,170 @@ import {
...
+            id: expect.any(String),
+          },
+        ],
+      },
+      headers: [
+        {
+          id: expect.any(String),
+          schema: {
+            'x-stoplight-id': expect.any(String),
+          },
```
marbemac, line 155: Hmm, is there value in having the empty schema on all the params? Does empty schema basically === no schema?

src/oas2/transformers/params.ts
```diff
@@ -157,10 +163,11 @@ export const translateFromFormDataParameters = withContext<
 
   return parameters.reduce((body, parameter) => {
     const { schema, description } = buildSchemaForParameter.call(this, parameter);
-    (body.contents || []).forEach(content => {
+    for (const content of Array.isArray(body.contents) ? body.contents : []) {
       delete schema.$schema;
+      delete schema['x-stoplight-id'];
 
-      if (description) {
+      if (typeof description === 'string') {
```
marbemac, line 170: `if (typeof description === 'string' && description) {`? Rather than adding empty description props basically.

P0lip: yeah, makes sense

src/oas2/transformers/__tests__/request.test.ts
P0lip: Yeah, true. I didn't touch that logic, but I agree the way it works is suboptimal.

marbemac (approved):

PR #183, opened 2022-04-15
Change JSON schema generator to genson-js

strmer15:
## Motivation and Context
The `json-schema-generator` dependency is not maintained anymore - it's been out of date for years, but it also has numerous security issues with its dependencies. There is another library, `gensons-js`, that is written in Typescript and is better maintained currently.

See https://github.com/stoplightio/prism/pull/2035 for an example of a dependency audit that can't fully resolve the security issues because of the version of `optimist` that `json-schema-generator` uses.

https://github.com/aspecto-io/genson-js

## Description
This change replaces the import and call to `jsonSchemaGenerator` with the `createSchema` function from `genson-js`. The main difference I found when testing is that when a string type is generated in the schema, there is no longer a `"minLength": 1` property added, meaning that empty strings are allowed in the schema. That doesn't seem like a huge issue, but it is probably a breaking change.

In addition, the types in the `convertSchema` function and associated functions used `OASSchemaObject` even though they would cast to `JSONSchema7`, so they were updated to use `OASSchemaObject | JSONSchema7`

## How Has This Been Tested?
Ran `yarn install`, `yarn build`, then `yarn test` and all the existing tests passed.

## Types of changes
- [ x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ x ] This PR's code follows as closely as possible [the coding style/guidelines of this project](???).
- [ x ] All new and existing tests pass locally (excluding flaky CI tests).

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "@types/swagger-schema-official": "~2.0.21",
     "@types/type-is": "^1.6.3",
     "fnv-plus": "^1.3.1",
-    "json-schema-generator": "^2.0.6",
+    "genson-js": "^0.0.8",
     "lodash.isequalwith": "^4.4.0",
     "lodash.pick": "^4.4.0",
     "lodash.pickby": "^4.6.0",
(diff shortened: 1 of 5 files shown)


strmer15:
Looks like `json-schema-generator` got forked to `@stoplight/json-schema-generator`, see https://github.com/stoplightio/json-schema-generator. Closing this as that resolves this problem.

PR #194, opened 2022-06-10
feat(oas): add bundleOas{2,3}Service functions

P0lip:
Needs: https://github.com/stoplightio/types/pull/110


marbemac (approved):
This looks good to me - phew what an improvement the bundle approach is!

Only thing is whether we want to pull the id gen functions along lines of https://github.com/stoplightio/platform-internal/blob/46e0f37d36bdab79f28c11b078157fc8384efda7/libs/stoplight-doc/src/import/ids.ts over to http-spec, to try to decrease possibility of mistake in the future. But could always do this in a separate PR if you prefer.

P0lip:
Yeah, I'm in for pulling them in. I like them being in one spot

PR #198, opened 2022-06-22
feat: move id generators

P0lip:


src/oas/__tests__/__snapshots__/operation.test.ts.snap
```diff
@@ -337,13 +337,13 @@ Array [
...
               "x-stoplight": Object {
-                "id": "3c572d09f18de",
+                "id": "8a951cfda4990",
               },
             },
             "style": "simple",
           },
         ],
-        "id": "ae2d0e59a4075",
+        "id": "ee11fabc87ff7",
```
P0lip, line 346: this one is due to `produces` being empty
Before the change we did

```ts
hash(`http_response-${this.parentId}-${actualKey}-${sortProducesOrConsumes(produces).join('-')}`)
```

and right now we do

```ts
hash([
      'http_response',
      props.parentId,
      props.codeOrKey,
      ...(Array.isArray(props.produces) ? sortAlphabetically(props.produces) : []),
]);
```

In the first case, we'll always have the trailing dash, but in the latter scenario that dash won't be there, hence the change in the ID.

src/oas/__tests__/__snapshots__/operation.test.ts.snap
```diff
@@ -315,20 +315,20 @@ Array [
         "headers": Array [
           Object {
             "description": "Allowed clients",
-            "id": "b1cba0180c7df",
+            "id": "f06fd57e462c1",
```
P0lip, line 318: and these all changes because the operation id is different

src/generators.ts
```diff
@@ -0,0 +1,87 @@
+type Context = { parentId: string };
+
+function hash(dependencies: (string | undefined)[]) {
+  return dependencies.join('-');
```
P0lip, line 4: I skipped the `hash` call here. We'll need to do

```
import { hash } from '@stoplight/http-spec/hash';

hash(idGenerators[kind](params))
```

This is because `http-spec` lets you provide a custom `generateId`. If we kept it hardcoded, we could call `hash` everywhere

P0lip: whatcha think @marbemac?

src/context.ts
```diff
@@ -9,12 +10,31 @@ import type {
   TranslateFunction,
 } from './types';
 
+function wrapGenerateId<T extends Record<string, unknown>>(
```
P0lip, line 13: this is so we don't need to pass `parentId` in `http-spec`.
I'm indifferent and if this is complicated, we can just pass `parentId` explicitly.

src/generators.ts
marbemac: > This is because http-spec lets you provide a custom generateId

I forget, did we end up using this functionality anywhere?

src/generators.ts
```diff
@@ -0,0 +1,87 @@
...
+function sanitizePath(path: string) {
+  return path.replace(/({)[^}]+(?=})/g, '$1');
+}
+
+export const idGenerators = {
+  tag: (props: Context & { name: string }) => {
+    return join(['tag', props.name]);
+  },
+
+  schema: (props: Context & { key: string }) => {
```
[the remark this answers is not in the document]
P0lip: Yeah, I've yalced it an updated the usage here https://github.com/stoplightio/platform-internal/pull/11931/commits/202fe28132329986481e0d3657ae39089dcf1207

src/oas2/transformers/servers.ts
```diff
@@ -49,7 +49,7 @@ export const translateToServer = withContext<
   const url = uri.toString().replace(/\/$/, ''); // Remove trailing slash
 
   return {
-    id: this.generateId(`http_server-${this.parentId}-${url}`),
+    id: this.generateId.httpServer({ url }),
```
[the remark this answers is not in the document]
P0lip: It's optional. Will remove it
This is mostly when we want to enforce we don't care about the placement / usage of a given object, but want to always scope it to the service.

PR #200, opened 2022-06-28
fix(oas2): do not bundle formData and body parameters

P0lip:
Fixes https://github.com/stoplightio/platform-internal/issues/12013

### DIFF
--- a/src/oas2/__tests__/bundle.test.ts
+++ b/src/oas2/__tests__/bundle.test.ts
@@ -15,6 +15,7 @@ describe('bundleOas2Service', () => {
           paths: {
             '/users/{userId}': {
               patch: {
+                consumes: ['application/json'],
                 parameters: [
                   {
                     $ref: '#/parameters/ids',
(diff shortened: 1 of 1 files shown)


src/oas2/__tests__/bundle.test.ts
```diff
@@ -4,6 +4,147 @@ import * as path from 'path';
...
+          id: 'c911d0f53d100',
+          method: 'patch',
+          path: '/users/{userId}',
+          extensions: {},
+          request: {
+            body: {
+              id: '9b5d65cfbc4fd',
+              contents: [],
+              required: true,
+            },
```
marbemac, line 81: Hmm I'm a little confused - where is the info from those shared parameters? Haven't we lost information in this bundled object versus the original openapi spec?

P0lip: We don't bundle them, yeah, so we lost one

P0lip: The thing is we need `consumes` in order to build a request payload.
Unlike OAS 3.x, request bodies in v2 aren't "independent".

> There is no information in the output object about the request body

because I didn't add any `consumes`, that's how http-spec was apparently implemented work for OAS 2.0.
If there's no `consumes`, we don't do anything meaningful 🤷‍♂️

marbemac: Ah ok - so if there is a consumes at top level of openapi, or on the specific operation, the data would be for the most part preserved in the output?

marbemac (approved):

PR #211, opened 2023-02-01
fix: add translate callback test

brendarearden:
## Motivation and Context
Adds test to cover callbacks translation to help with tracking down the root cause for the following issue: 
[#15026](https://github.com/stoplightio/platform-internal/issues/15026)

## Description
Added a very basic test

## Types of changes

## Checklist

### DIFF
--- a/src/oas3/transformers/__tests__/callbacks.test.ts
+++ b/src/oas3/transformers/__tests__/callbacks.test.ts
@@ -1,0 +1,252 @@
+import { DeepPartial } from '@stoplight/types';
+import { OpenAPIObject } from 'openapi3-ts';
+
+import { createContext } from '../../../oas/context';
+import { translateToCallbacks as _translateToCallbacks } from '../callbacks';
+
+const translateToCallbacks = (document: DeepPartial<OpenAPIObject>, callbacks: unknown) =>
+  _translateToCallbacks.call(createContext(document), callbacks);
+
+describe('translateToCallbacks', () => {
+  it('given multiple callback entries, it should translate', () => {
+    const callbackEntries = {
+      newPetWebhook: {
+        '{$request.body#/newPetAvailableUrl}': {
+          post: {
+            requestBody: {
+              required: true,
+              content: {
+                'application/json': {
+                  schema: {
+                    type: 'object',
+                    properties: {
+                      message: {
+                        type: 'string',
+                        example: 'A new pet has arrived',
+                      },
+                    },
+                    required: ['message'],
+                  },
+                },
+              },
+            },
+            responses: {
+              '200': {
+                description: 'Your server returns this code if it accepts the callback',
+              },
+            },
+          },
+        },
+        '{$request.body#/returnedPetAvailableUrl}': {
+          post: {
+            requestBody: {
+              required: true,
+              content: {
+                'application/json': {
+                  schema: {
+                    type: 'object',
+                    properties: {
+                      message: {
+                        type: 'string',
+                        example: 'A pet has been returned',
+                      },
+                    },
+                    required: ['message'],
+                  },
+                },
+              },
+            },
+            responses: {
+              '200': {
+                description: 'Your server returns this code if it accepts the callback',
+              },
+            },
+          },
+        },
+      },
+      petAdopted: {
+        '{$request.body#/adoptedUrl}': {
+          post: {
+            requestBody: {
+              required: true,
+              content: {
+                'application/json': {
+                  schema: {
+                    type: 'object',
+                    properties: {
+                      message: {
+                        type: 'string',
+                        example: 'A pet has been adopted',
+                      },
+                    },
+                    required: ['message'],
+                  },
+                },
+              },
+            },
+            responses: {
+              '200': {
+                description: 'Your server returns this code if it accepts the callback',
+              },
+            },
+          },
+        },
+      },
+    };
+    expect(translateToCallbacks({}, callbackEntries)).toStrictEqual([
+      {
+        callbackName: 'newPetWebhook',
+        extensions: {},
+        id: expect.any(String),
+        method: 'post',
+        path: '{$request.body#/newPetAvailableUrl}',
+        request: {
+          body: {
+            contents: [
+              {
+                encodings: [],
+                examples: [],
+                id: expect.any(String),
+                mediaType: 'application/json',
+                schema: {
+                  $schema: 'http://json-schema.org/draft-07/schema#',
+                  properties: {
+                    message: {
+                      examples: ['A new pet has arrived'],
+                      type: 'string',
+                    },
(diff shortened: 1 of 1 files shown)


daniel-white (commented):
seems reasonable to me. of course i'd defer some to @P0lip

daniel-white (approved):

src/generators.ts
```diff
@@ -37,6 +37,10 @@ export const idGenerators = {
     return join(['http_operation', props.parentId, props.method, sanitizePath(props.path)]);
   },
 
+  httpCallbackOperation: (props: Context & { method: string; path: string }) => {
+    return join(['http_operation', props.parentId, props.method, props.path]);
```
P0lip, line 41:
```suggestion
    return join(['http_callback', props.parentId, props.method, props.path]);
```

src/context.ts
```diff
@@ -66,6 +66,7 @@ export function createContext<T extends Record<string, unknown>>(
       service: '',
       path: '',
       operation: '',
+      callback: '',
```
P0lip, line 69:
```suggestion
```
not used

src/oas3/transformers/callbacks.ts
```diff
@@ -19,11 +20,14 @@ export const translateToCallbacks: Oas3TranslateFunction<[callbacks: unknown], I
             paths: { [path]: { [method]: op } },
           };
 
+          const ctx = createContext(document);
+          ctx.context = 'callback';
```
P0lip, line 24: you need to copy over ids.
```suggestion
          Object.assign(ctx.ids, this.ids);
```

src/__tests__/generators.test.ts
```diff
@@ -0,0 +1,31 @@
+import { idGenerators } from '../generators';
+
+describe('idGenerators', () => {
```
P0lip, line 3: I wouldn't recommend testing it this way.
For instance, in this case, the IDs wouldn't be unique because the ctx does not have correct IDs.
You can toss in a spec here src/oas3/__tests__/__fixtures__

P0lip (changes requested):

src/oas/operation.ts
```diff
@@ -57,7 +57,16 @@ export const transformOasOperation: TranslateFunction<
 
   const serviceId = (this.ids.service = String(this.document['x-stoplight']?.id));
   this.ids.path = this.generateId.httpPath({ parentId: serviceId, path });
```
P0lip, line 59:
```suggestion
  const serviceId = (this.ids.service ||= String(this.document['x-stoplight']?.id));
  this.ids.path ||= this.generateId.httpPath({ parentId: serviceId, path });
```
perhaps make it optional so that they don't get re-assigned

brendarearden: Changing this resulted in re-assignment not working when it was supposed to (almost all the tests started failing with the incorrect serviceId etc). I will not be making this change as part of the PR - should I open an issue that covers this?

src/__tests__/generators.test.ts
brendarearden: I am able to confirm the ids are unique in the traverse callback tests I added, so these tests were mainly added to confirm string output from the idGenerator was sanitizing the path still for http_operations and not for http_callbacks. 

The way I understood the id generator to be working was the string from the idGenerator is used as part of generating the actual id on the TransformerContext. So if the string value is unique, the generated id should be unique because its what gets hashed. Is this accurate?

src/oas3/transformers/callbacks.ts
brendarearden: I added `Object.assign(ctx.ids, this.ids);` but I also retained `ctx.context = 'callback';`

src/generators.ts
brendarearden: good catch! it has been updated in the latest commit

src/context.ts
brendarearden: I removed this, but since I had updated the type, I had to change to use https://github.com/stoplightio/http-spec/pull/211/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR42 and https://github.com/stoplightio/http-spec/pull/211/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR41 to ensure the types resolved properly.

src/__tests__/generators.test.ts
P0lip: >So if the string value is unique, the generated id should be unique because its what gets hashed.

the problem is we don't have a test that'd verify whether the `parentId` actually gets set, and if `parentId` is not set properly, the IDs may no longer be stable if one of the dependencies is always set to `undefined`.
That's why I suggested adding a document do that ids directory.

src/oas/operation.ts
P0lip: ah, I know why the tests fail.
There's no bug in the code, the bug would be actually introduced if you applied my suggestion 😆

src/types.ts
```diff
@@ -31,15 +31,15 @@ export type HttpOperationTransformer<T> = (opts: T) => IHttpOperation;
...
 
 export type References = Record<string, { resolved: boolean; value: string }>;
 
 export type TransformerContext<T extends Fragment = Fragment> = {
   document: T;
   context: AvailableContext;
-  parentId: string;
-  readonly ids: Record<AvailableContext, string>;
+  parentId: string | undefined;
+  readonly ids: Partial<Record<AvailableContext, string>>;
```
P0lip, line 42:
```suggestion
  readonly ids: Omit<Record<AvailableContext, string>, 'callback'>;
```
perhaps we do it this way to better reflect the actual value ids may hold?

src/types.ts
```diff
@@ -31,15 +31,15 @@ export type HttpOperationTransformer<T> = (opts: T) => IHttpOperation;
...
+export type AvailableContext = 'service' | 'path' | 'operation' | 'callback';
 
 export type References = Record<string, { resolved: boolean; value: string }>;
 
 export type TransformerContext<T extends Fragment = Fragment> = {
   document: T;
   context: AvailableContext;
-  parentId: string;
-  readonly ids: Record<AvailableContext, string>;
+  parentId: string | undefined;
```
P0lip, line 41:
```suggestion
  parentId: string;
```
this may be no longer needed if we do the omit part

src/oas3/transformers/callbacks.ts
brendarearden: I went ahead and updated to exactly you suggestion in the meantime

src/context.ts
brendarearden: this was still removed as part of switching to the omit pattern

P0lip (approved):

PR #230, opened 2023-05-05
feat(oas3): expose explicit props and allow props in params

daniel-white:
additional metadata needed for oas3 - automerge is on!


brendarearden (approved):
👍

PR #235, opened 2023-05-23
fix(oas2): include shared responses in bundled service

P0lip:
Even though they aren't used directly in the bundle itself (due to our handling of consumes/produces), we should include them so that SL Doc can correctly link cross-file references.


billiegoose (approved):
lol

PR #251, opened 2023-08-28
feat(oas): support info extensions

daniel-white:
collects extensions in the `info` section as `infoExtensions`

### DIFF
--- a/src/oas/service.ts
+++ b/src/oas/service.ts
@@ -49,6 +49,7 @@ export const transformOasService: TranslateFunction<DeepPartial<OpenAPIObject> |
 
       ...toExternalDocs(document.externalDocs),
       extensions: getExtensions(document),
+      infoExtensions: getExtensions(document.info),
     };
 
     if (isPlainObject(document.info) && hasXLogo(document.info)) {
--- a/src/oas3/__tests__/service.test.ts
+++ b/src/oas3/__tests__/service.test.ts
@@ -46,6 +46,7 @@ describe('oas3 service', () => {
           id: 'abc',
         },
       },
+      infoExtensions: {},
     });
   });
 
@@ -104,6 +105,7 @@ describe('oas3 service', () => {
           id: 'abc',
         },
       },
+      infoExtensions: {},
     });
   });
 
@@ -121,6 +123,7 @@ describe('oas3 service', () => {
           id: 'abc',
         },
       },
+      infoExtensions: {},
     });
   });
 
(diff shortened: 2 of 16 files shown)


src/oas/service.ts
```diff
@@ -49,6 +49,7 @@ export const transformOasService: TranslateFunction<DeepPartial<OpenAPIObject> |
 
       ...toExternalDocs(document.externalDocs),
       extensions: getExtensions(document),
+      infoExtensions: getExtensions(document.info),
```
EdVinyard, line 52: Is the additional property `infoExtensions` needed because there's no node for `info` itself, so we can't distinguish between `x-widget` at the "top level" and `x-widget` nested under `info`?  

Looking past this ticket, are there any other [children of the top-level OpenAPI object](https://spec.openapis.org/oas/v3.0.1.html#fixed-fields) that would benefit from similar treatment, or even a more general solution?

src/oas3/__tests__/service.test.ts
```diff
@@ -412,6 +429,7 @@ describe('oas3 service', () => {
           altText: 'logo',
           href: 'https://stoplight.io',
         },
+        infoExtensions: { 'x-logo': {} },
```
EdVinyard, line 432: @matthewmurphy , I think **nine** of the files changed in this PR are solely because we're doing snapshot or test-the-whole-result style testing.  I wonder when we should consider making some of these tests in `http-spec` more targeted.  🤔

EdVinyard (commented):
I have a couple questions I hope you'll consider, but everything looks good to me.

EdVinyard (approved):

src/oas/service.ts
daniel-white: @EdVinyard  correct. a more general solution would likely require a larger rework of how extensions are stored

src/oas3/__tests__/service.test.ts
daniel-white: a snapshot test would have been great bc i could have had jest update them after justifying the change

src/oas/service.ts
EdVinyard: Are there any other [children of the top-level OpenAPI object](https://spec.openapis.org/oas/v3.0.1.html#fixed-fields) that would benefit from similar treatment, or even a more general solution?  That is, should I write a follow-up ticket to fix other related errors that won't be fixed by this PR?

PR #255, opened 2023-11-29
Feat(oas3.0): add nullable to explicit properties

kaylachun:
## Motivation and Context

Addresses: https://github.com/stoplightio/platform-internal/issues/18151

## Types of changes

## Checklist

### DIFF
--- a/src/oas/transformers/schema/__tests__/schema.spec.ts
+++ b/src/oas/transformers/schema/__tests__/schema.spec.ts
@@ -294,7 +294,6 @@ describe('translateSchemaObject', () => {
         },
         $schema: 'http://json-schema.org/draft-07/schema#',
         type: ['string', 'null'],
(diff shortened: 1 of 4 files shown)


kaylachun:
> The idea behind this schema converter is to turn any OAS 2.0 & 3.0.x Schema Objects into valid JSON Schema (draft 7 in our case), so that any consumer of http-spec can safely assume they don't need to account for various shapes. Doing anything based on the provided spec kind of undermines the purpose of the converter since you'll now have different shapes depending on the provided spec version. Overall though, in SL, it should be relatively easy to tell whether `nullable` was set in the first place or not, since `type` would contain `null`.

That makes sense I figured it might be the wrong approach to be checking for the oas version here.  I'll try to infer nullable if type includes null in sl-doc then.  Thank you!

PR #258, opened 2023-12-11
feat(oas3): adds webhook support for oas 3.1

daniel-white:
this is a set of changes to support oas 3.1 webhooks. this does have breaking changes to make the operation parsing compatible both with regular operations and webhooks.

### DIFF
--- a/src/oas2/__tests__/operation.test.ts
+++ b/src/oas2/__tests__/operation.test.ts
@@ -1,5 +1,4 @@
 import { DeepPartial } from '@stoplight/types';
-import { OpenAPIObject } from 'openapi3-ts';
 import { Spec } from 'swagger-schema-official';
 
 import { setSkipHashing } from '../../hash';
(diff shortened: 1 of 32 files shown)


src/oas2/__tests__/operation.test.ts
```diff
@@ -226,7 +229,7 @@ describe('transformOas2Operation', () => {
   });
 
   it('given malformed parameters should translate operation with those parameters', () => {
-    const document: Partial<OpenAPIObject> = {
+    const document: Partial<DeepPartial<Spec>> = {
```
matthewmurphy, line 232: Does this need the `Partial`? Wouldn't this be the same as just `DeepPartial<Spec>`?

daniel-white: i can fix that

matthewmurphy (approved):

src/postman/types.ts
```diff
@@ -1,9 +1,13 @@
...
 
-export type PostmanCollectionHttpOperationTransformer = HttpOperationTransformer<{
-  document: CollectionDefinition;
-  path: string;
-  method: string;
-}>;
+export type PostmanCollectionHttpOperationTransformer = HttpEndpointOperationTransformer<
+  {
+    document: CollectionDefinition;
+    path: string;
```
brendarearden, line 9: is this supposed to be `name: string;` and not `path: string;`

daniel-white: no this one should be fine

brendarearden (approved):
looks good!

P0lip (approved):

PR #262, opened 2023-12-20
fix: callback operation ids

mallachari:
Addresses: https://github.com/stoplightio/platform-internal/issues/18498

## Motivation and Context
Callback ids are currently being generated based on operation method and path. This may lead to duplicated ids in case where one operation has more than one callbacks defined.

## Description
Adds additional key property (callback name) when generating ids for callback nodes.

## How Has This Been Tested?
Tests updated

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/src/generators.ts
+++ b/src/generators.ts
@@ -48,8 +48,8 @@ export const idGenerators = {
     return join(['http_webhook_operation', props.parentId, props.method, sanitizePath(props.name)]);
   },
 
-  httpCallbackOperation: (props: Context & { method: string; path: string }) => {
-    return join(['http_callback', props.parentId, props.method, props.path]);
+  httpCallbackOperation: (props: Context & { method: string; path: string; key: string }) => {
+    return join(['http_callback', props.parentId, props.method, props.path, props.key]);
   },
 
   httpPathParam: (props: Context & { keyOrName: string }) => {
--- a/src/generators.ts
+++ b/src/generators.ts
@@ -49,7 +49,7 @@ export const idGenerators = {
   },
 
   httpCallbackOperation: (props: Context & { method: string; path: string; key: string }) => {
-    return join(['http_callback', props.parentId, props.method, props.path, props.key]);
+    return join(['http_callback', props.parentId, props.key, props.method, props.path]);
   },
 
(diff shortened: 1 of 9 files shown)


src/oas/operation.ts
```diff
@@ -89,6 +89,7 @@ export const transformOasEndpointOperation: TranslateFunction<
         parentId: serviceId,
         method,
         path: name,
+        key: key ?? 'unknown',
```
mallachari, line 92: Or should we generate random string in that case?

src/types.ts
```diff
@@ -36,6 +36,7 @@ export interface ITransformEndpointOperationOpts<T extends Fragment> {
   document: T;
   name: string;
   method: string;
+  key?: string;
```
mallachari, line 39: Had to add it because `transformOas3Operation` [uses](https://github.com/stoplightio/http-spec/pull/262/files#diff-edf0102d749d7bcfcaf6f168492e0d23f08a03740e99332863026068cccc3feeR47) this type

src/generators.ts
```diff
@@ -48,8 +48,8 @@ export const idGenerators = {
     return join(['http_webhook_operation', props.parentId, props.method, sanitizePath(props.name)]);
   },
 
-  httpCallbackOperation: (props: Context & { method: string; path: string }) => {
-    return join(['http_callback', props.parentId, props.method, props.path]);
+  httpCallbackOperation: (props: Context & { method: string; path: string; key: string }) => {
+    return join(['http_callback', props.parentId, props.method, props.path, props.key]);
```
daniel-white, line 52: lets flip this around so that its between parent and method

src/oas/operation.ts
daniel-white: i think it would have to be something stable. i'm kind of thinking this should throw but @P0lip might have a suggestion

P0lip: You can leave an empty key. That's what we do for schemas

mallachari: Updated

daniel-white (approved):

PR #267, opened 2024-04-30
fix(oas): extensions on tags is incorrectly spread

P0lip:
`INodeTag` extends `ISpecExtensions` and as such should have `extensions` property containing all collected spec extensions.
We accidentally spread `extensions` causing all spec extensions to be placed among other INodeTag members.


weyert:
Would it be possible to merge this? Looks useful now that Elements supports vendor extensions

PR #268, opened 2024-06-17
feat(oas3): return security scopes for types other than oauth2

darekplawecki:
## Motivation and Context
The PR is required for displaying security scheme roles in elements: https://github.com/stoplightio/elements/issues/2491

## Description
The `scopes` value has been added to `extensions` as it's not part of the security scheme, so I didn't want to extend `HttpSecurityScheme`.

## How Has This Been Tested?
Tested locally in elements.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] This PR's code follows as closely as possible [the coding style/guidelines of this project](???).
- [x] All new and existing tests pass locally (excluding flaky CI tests).

### DIFF
--- a/src/oas3/accessors.ts
+++ b/src/oas3/accessors.ts
@@ -45,11 +45,14 @@ export function getSecurities(
           ];
         }
 
+        const extensions = scopes?.length ? { ['x-scopes']: scopes } : {};
+
         return [
           opScheme,
           {
             ...definition,
-            extensions: getExtensions(definition),
+            ...extensions,
+            extensions: getExtensions({ ...definition, ...extensions }),
           },
         ];
       })
--- a/src/oas3/__tests__/accessors.test.ts
+++ b/src/oas3/__tests__/accessors.test.ts
@@ -1,4 +1,5 @@
 import { DeepPartial } from '@stoplight/types';
+import { SecuritySchemeType } from 'openapi3-ts';
 
 import { setSkipHashing } from '../../hash';
 import { getSecurities as _getSecurities, OperationSecurities } from '../accessors';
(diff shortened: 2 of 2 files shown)


src/oas3/accessors.ts
```diff
@@ -45,11 +45,14 @@ export function getSecurities(
...
         }
 
+        const extensions = scopes?.length ? { ['x-scopes']: scopes } : {};
+
         return [
           opScheme,
           {
             ...definition,
-            extensions: getExtensions(definition),
+            ...extensions,
```
mallachari, line 54: Do we need all extensions to be listed here? Can't we rely just on `extensions` object?

mallachari (commented):
I think that's good choice putting it into extensions since it is not part of the specification. Just one question regarding returned data.

src/oas3/accessors.ts
darekplawecki: That's what I thought at the beginning, but it turned out the `extensions` object is completely ignored in the `securities.ts#translateToSingleSecurity` method and calculated again from the `securityScheme` object: 
```ts
  ...
  const baseObject: { id: string; key: string; description?: string; extensions: Extensions } = {
    id: this.generateId.httpSecurity({
      keyOrName: key,
      kind,
      ...(kind === 'requirement'
        ? {
            index,
            scopeKeys: getScopeKeys(securityScheme),
          }
        : {}),
    }),
    key,
    extensions: getExtensions(securityScheme),
  };
  ...
```
We could simply do `extensions: securityScheme.extensions`, but I wasn't sure where else is the method used and if it breaks something else.

lukasikp (approved):

src/oas3/accessors.ts
mallachari: Yea, you're right about that, we're loosing these extensions. Maybe we could try to include it in there with `extensions: getExtensions({ ...securityScheme, ...securityScheme.extensions })`, so we keep both, but I'm ok with your current approach too.

mallachari (approved):

